PR #57, opened 2018-03-10
Ops/nvmrc

WesleyCharlesBlake:
I included an .nvmrc file for development purpose, to ensure the correct version of nodejs is installed with nvm


README.md
```diff
@@ -38,8 +38,8 @@ curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | b
 # Activate NVM
 . ~/.nvm/nvm.sh
 
-# Install NodeJS
-nvm install 9.3.0
+# Install NodeJS (will install node version in .nvmrc)
```
lautarodragan: Would this only happen if the `nvm install` was after the `git clone` and `cd node`?

WesleyCharlesBlake: @lautarodragan yes, so nvm will check for a local version of node specified by the `.nvmrc` and it will use that. Issuing the `nvm install` is idempotent, so we can issue it multiple times, and achieve the same result, without having to downloading the nvm node packages multiple times if it exists locally. It will need to be done from the same directory as the .nvmrc file. I have updated the readme! Sorry about that!

WesleyCharlesBlake:
@lautarodragan I will handle this better with the Make file refactor branch

PR #65, opened 2018-04-09
Pino JS

### LINKED ISSUE #16: PinoJS
Drop all console.log / console.error in favor of PinoJS.

All logging must be structured.

Define a basic loggin structure structure. Current candidate is:

```
{
  "module": "Storage",
  "file": "ClaimController",
  "method": "downloadNextHash",
  "message": "A more human-readable message describing what's going on",
  extraVariables,
  alwaysUsingShorthandPropertyNames,
}
```

### LINKED ISSUE #54: Broken Exception Logging
Logging of exceptions is currently broken at at least one place:

https://github.com/poetapp/node/blob/aad2f4804de85646875257158d1ea465a03af030/src/Storage/ClaimController.ts#L92-L109

`JSON.stringify(new Error('123'))` always returns an stringified empty object.

A quick test with PinoJS didn't seem to solve the issue, although further testing & researching should be done (I'd expect it to support this use case out of the box).

If PinoJS doesn't support this use case out of the box, [this](https://stackoverflow.com/a/18391400/1524406) seems to be an acceptable solution.

Needs #16 (could possibly solve this out of the box)

lautarodragan:
This PR replaces all `console.log` with PinoJS logging.

It also introduces more consistency in the logging messages and structures, and now correctly logs Errors.

Left out of this PR were the `console.log` calls in Messaging. This file will [undergo a refactor](https://github.com/poetapp/node/issues/66) which shall include this change.

Fixes #16 
Fixes #54


src/BlockchainReader/BlockchainReaderService.ts
```diff
@@ -1,23 +1,30 @@
...
+    @inject('Logger') logger: Pino.Logger,
     @inject('ClaimController') claimController: ClaimController,
-    @inject('BlockchainReaderServiceConfiguration') configuration: BlockchainReaderServiceConfiguration
+    @inject('BlockchainReaderServiceConfiguration') configuration: BlockchainReaderServiceConfiguration,
   ) {
+    this.logger = childWithFileName(logger, __filename)
     this.claimController = claimController
     this.configuration = configuration
     this.interval = new Interval(this.scanNewBlock, this.configuration.blockchainReaderIntervalInSeconds * 1000)
+    this.logger.info('wachin')
```
wzalazar: What means ```wachin``` ?

src/Helpers/Logging.ts
```diff
@@ -0,0 +1,20 @@
+import { basename } from 'path'
+import * as Pino from 'pino'
+import { LoggingConfiguration } from '../Configuration'
+
+export function newModuleLogger(configuration: LoggingConfiguration, dirname: string): Pino.Logger {
```
wzalazar: I dislike the prefix ```new```,  just it's my point of view.

src/Storage/Storage.ts
```diff
@@ -22,10 +25,11 @@ export class Storage {
 
   constructor(configuration: StorageConfiguration) {
     this.configuration = configuration
+    this.logger = newModuleLogger(configuration, __dirname)
   }
 
   async start() {
-    console.log('Storage Starting...', this.configuration)
+    this.logger.info({ configuration: this.configuration }, 'Starting Starting')
```
wzalazar: twice Starting Starting ?

PR #71, opened 2018-04-18
adds user overrides for integration test config

### LINKED ISSUE #39: Integration Testing: Configurable Node URL
Allow setting the URL of The Po.et Node for the integration tests via environment variables. 

`export POET_INTEGRATION_NODE_URL="..."`

Keep `http://localhost:18080` as the default.

warrenv:
Previously, the mongo database url and node integration test url were
hard-coded.

This commit sets the mongo db url from the app configuration. It also
allows the integration test url to be overridden by setting the
POET_INTEGRATION_NODE_URL environment variable.

Fixes #39


test/PrepareDB.ts
```diff
@@ -1,14 +1,16 @@
 import { Db, MongoClient } from 'mongodb'
 
+import { loadConfigurationWithDefaults } from 'Configuration'
```
lautarodragan: Avoid importing things of /src from /test. There's no guarantee that `POET_INTEGRATION_NODE_URL` points to that same node.

warrenv: ok, I've added POET_MONGODB_URL.

.env.example
```diff
@@ -0,0 +1,3 @@
+# override the default integration test url.
```
lautarodragan: Let's either place this file in /test if possible (I think not, not sure) or make it very clear these env vars are for testing.

warrenv: I feel like the way we are handling env vars goes against 12 factor app best practices by using a config file in the user's home directory. It makes sense when just looking at a developer's environment, but I don't feel like this is going to be scalable across additional environments (staging, prod, etc) In a container based environment, the env variables need to be passed in via docker.

I see an approach that uses .env to hold the overrides that are currently come from ~/.po.et/configuration. And then .env.test would contain those plus the additional env vars needed for testing.

cc: @WesleyCharlesBlake

PR #78, opened 2018-04-24
updates .travisci with lint and test

warrenv:
Previously, travisCI was not configured to run linting or tests.

This commit updates the .travis.yml file and adds a .nvmrc file so TravisCI
will use our configured nodejs version.

NOTE: linting is disabled for now until failures are fixed.


.travis.yml
```diff
@@ -14,3 +12,6 @@ before_script:
   - sleep 5
   - npm start &
   - sleep 5
+script:
+  # - npm run lint
+  - npm run test
```
wzalazar: I don't sure we have to execute this test command because we have 3 commands now.

```
"test": "nyc alsatian './dist/babel/test/Integration/**/*.js'",
"test-unit": "nyc alsatian './dist/babel/test/Unit/**/*.js'",
"test-integration": "nyc alsatian './dist/babel/test/Integration/**/*.js'",
```

`test` are equal to `test-integration`

What command is right @lautarodragan?

lautarodragan: That's some legacy stuff there.

`npm test` is the only valid one currently. The other too were relevant in the past, but no longer.

PR #87, opened 2018-05-03
fixes lint failures

warrenv:
Previously there were many lint failures after refactoring to use the
@po.et/tlsint-rules npm package.

This commit fixes all lint failures and re-enables linting on Travis.


src/BlockchainReader/BlockchainReaderService.ts
```diff
@@ -18,17 +18,24 @@ export class BlockchainReaderService {
   constructor(
     @inject('Logger') logger: Pino.Logger,
     @inject('ClaimController') claimController: ClaimController,
-    @inject('BlockchainReaderServiceConfiguration') configuration: BlockchainReaderServiceConfiguration,
+    @inject('BlockchainReaderServiceConfiguration')
```
lautarodragan: Why is this line being broken into two? Exceeded max char count?

lautarodragan: So this line looks like it's 104 characters long. The rule is set for a max line length of 120. Wonder what's going on.

warrenv: That was probably prettier doing it for some reason.

wzalazar: yes, there is an issue in the rules.  I added the task ```tslint-config-prettier-check``` in the package and when execute the output is:

Conflict rule(s) detected in ./poet-tslint.json:
  indent
  max-line-length

For resolve it, we have to remove this lines

https://github.com/poetapp/tslint-rules/blob/86bd6b27a3fe589186a6dbdda51287eba2fe3d43/poet-tslint.json#L14

https://github.com/poetapp/tslint-rules/blob/86bd6b27a3fe589186a6dbdda51287eba2fe3d43/poet-tslint.json#L16

after that we have to add this rules here, 

https://github.com/poetapp/tslint-rules/blob/86bd6b27a3fe589186a6dbdda51287eba2fe3d43/poet-tslint.json#L33-L36

like as result:

```
     "prettier": [true, {
        "singleQuote": true,
        "semi": false,
        "printWidth": 120,
        "tabWidth": 2
      }],
```

lautarodragan: Perfect. Thanks @wzalazar ! Can you create an issue in the tslint-rules repo to track that?

We'll merge this as is for now.

wzalazar: ok, perfect!!!

PR #97, opened 2018-05-09
[Fixes #63] Improve Node Syncing

### LINKED ISSUE #63: Bug While Syncing Nodes
I found some works are not downloading correctly from IPFS.

They had `db.storage.attemps >= 5`, meaning they had been tried that many times and all failed. But manually resetting that value in the DB triggered another attempt, which succeeded immediately. 

We'll need to upgrade the retry system to have intervals between retries.

kennylavender:
Work In Progress. Do not Review or Merge. Code changes are very messy atm.

resolves #63 

- Increase timeout for claim downloads   
   - I noticed an incredible amount of claims failing because of the low timeout, upping this to 10 seconds improved the success rate. A larger claim download timeout can now be used because we have a much larger delay between attempts per hash.

### DIFF
--- a/src/Storage/ClaimController.ts
+++ b/src/Storage/ClaimController.ts
@@ -53,19 +53,24 @@ export class ClaimController {
     const logger = this.logger.child({ method: 'download' })
 
     logger.trace({ ipfsHashes }, 'Downloading Claims')
-
-    await this.collection.insertMany(
-      ipfsHashes.map(ipfsHash => ({ ipfsHash, claimId: null })),
-      { ordered: false }
-    )
+    try {
+      await this.collection.insertMany(
+        ipfsHashes.map(ipfsHash => ({ ipfsHash, claimId: null, lastDownloadAttempt: null, lastDownloadSuccess: null })),
+        { ordered: false }
+      )
+    } catch (e) {
+      logger.info('error inserting many')
+    }
   }
 
   async handleHashDownload(ipfsHash: string) {
     const logger = this.logger.child({ method: 'handleHashDownload' });
+    logger.info('downloading', ipfsHash)
 
     try {
       logger.trace({ ipfsHash }, 'Downloading Next Hash')
       const claim = await this.downloadClaim(ipfsHash)
+      logger.info('Successfully downloaded claim from IPFS');
       logger.info({ ipfsHash, claim }, 'Successfully downloaded claim from IPFS')
       await this.updateClaimIdIPFSHashPairs([
         {
@@ -81,6 +86,7 @@ export class ClaimController {
         }
       ])
     } catch (exception) {
+      logger.info('failed to download claim');
       logger.debug({ ipfsHash, exception }, 'Failed to Download Claim')
     }
   }
@@ -92,19 +98,72 @@ export class ClaimController {
     const logger = this.logger.child({ method: 'downloadNextHash' })
 
     logger.trace('Downloading Next Hash')
+    let record;
 
-    await this.collection.findOneAndUpdate({
-      claimId: null,
-      lastDownloadSuccess: { $exists: false },
-      lastDownloadAttempt: { $lt: currentTime + downloadDelay }
+    try {
+
+      
+      record = await this.collection.findOne({
+        claimId: null,
+        ipfsHash: { $exists: true },
+        $and: [
+          {
+            $or: [
+              { lastDownloadAttempt: null },
+              { lastDownloadAttempt: { $exists: false } },
+              { lastDownloadAttempt: { $lt: currentTime - downloadDelay } },
+            ]
+          },
+          {
+            $or: [
+              { lastDownloadSuccess: null },
+              { lastDownloadSuccess: { $exists: false } }
+            ]
+          }
+        ]
+      });
+    } catch(e) {
+      logger.info('failed to find a downloadable record');
+    }
+
+    if (!record) {
+      logger.info('no downloadable records found');
+      return;
+    }
+
+    logger.info(record);
+    try {
+    await this.collection.updateOne({
+      _id: record._id
     }, {
-      lastDownloadAttempt: currentTime
-    },
-    (error, { value }) => {
-      if (error) return;
-      if (!value || !value.ipfsHash) return;
-      this.handleHashDownload(value.ipfsHash);
-    })    
+      $set: { lastDownloadAttempt: currentTime }
+    });
+  } catch(e) {
+    logger.info('failed to update lastDownloadAttempt');
+  }
+
+    this.handleHashDownload(record.ipfsHash);
+    
+
+    // await this.collection.findOneAndUpdate({
+    //   claimId: null,
+    //   lastDownloadSuccess: { $exists: false },
+    //   ipfsHash: { $exists: true },
+    //   $or: [
+    //     { lastDownloadAttempt: { $lt: currentTime - downloadDelay } },
+    //     { lastDownloadAttempt: { $exists: false } }
+    //   ]
+    // }, {
+    //   lastDownloadAttempt: currentTime
+    // },
+    // (error, result) => {
+    //   logger.info(error, result);
+    //   if (error) return;
+    //   if (!result) return;
+    //   if (!result.value) return;
+    //   if (!result.value.ipfsHash) return;
+    //   
+    // })    
   }
 
   private downloadClaim = async (ipfsHash: string): Promise<Claim> => {
--- a/src/Storage/ClaimController.ts
+++ b/src/Storage/ClaimController.ts
@@ -60,47 +60,51 @@ export class ClaimController {
     )
   }
 
-  async downloadNextHash() {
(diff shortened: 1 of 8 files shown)


src/Storage/ClaimController.ts
```diff
@@ -53,51 +54,146 @@ export class ClaimController {
...
+    return {
+      entry,
+      currentTime,
+      ...rest,
+    }
+  }
+
+  async downloadNextHash({ retryDelay = 600000, maxAttempts = 20 }: { retryDelay: number; maxAttempts: number }) {
+    return this.findEntryToDownload({ retryDelay, maxAttempts })
+      .then(this.updateEntryAttempts.bind(this))
```
kennylavender: I am not accustomed to using classes, I need to see if I can structure this in a way that doesn't require bind.

kennylavender: I originally wanted to use ramda pipeP, but that doesn't seem to have typescript types.

kennylavender:
@lautarodragan ok, ready for review.

src/Storage/ClaimController.ts
```diff
@@ -53,51 +54,142 @@ export class ClaimController {
...
 
     logger.trace({ ipfsHashes }, 'Downloading Claims')
-
-    await this.collection.insertMany(ipfsHashes.map(ipfsHash => ({ ipfsHash, claimId: null })), { ordered: false })
+    await this.collection.insertMany(
+      ipfsHashes.map((ipfsHash): Entry => ({
+        ipfsHash,
+        claimId: null,
+        lastDownloadAttempt: null,
+        lastDownloadSuccess: null,
```
lautarodragan: Why `lastDownloadSuccess`? Can there be more than one download success?

src/Storage/ClaimController.ts
```diff
@@ -53,51 +54,142 @@ export class ClaimController {
     const logger = this.logger.child({ method: 'download' })
 
     logger.trace({ ipfsHashes }, 'Downloading Claims')
-
-    await this.collection.insertMany(ipfsHashes.map(ipfsHash => ({ ipfsHash, claimId: null })), { ordered: false })
+    await this.collection.insertMany(
+      ipfsHashes.map((ipfsHash): Entry => ({
```
lautarodragan: `ipfsHashes.map((ipfsHash): Entry` => ipfsHash is not of type `Entry` here, it's `string`. The type is inferred though, so you can just avoid declaring the type.

src/Storage/ClaimController.ts
```diff
@@ -53,51 +54,142 @@ export class ClaimController {
...
-      return
-    }
-
-    logger.info({ ipfsHash, claim }, 'Successfully downloaded claim from IPFS')
-
-    await this.updateClaimIdIPFSHashPairs([
+  async updateEntryPairs({ entry, claim, ...rest }: { claim: Claim; entry: Entry }) {
+    const logger = this.logger.child({ method: 'publishEntryDownload' })
+    logger.trace('started claim update hash pairs')
+    this.updateClaimIdIPFSHashPairs([
```
lautarodragan: Avoid ignoring returned promises.

If you don't `await` the `this.updateClaimIdIPFSHashPairs`, the promise returned by it will get ignored, which means that any error thrown inside `updateClaimIdIPFSHashPairs` will get ignored too.

src/Storage/ServiceConfiguration.ts
```diff
@@ -1,3 +1,5 @@
 export interface ServiceConfiguration {
   readonly downloadIntervalInSeconds: number
+  readonly downloadRetryDelayInMinutes: number
```
lautarodragan: Please create a ControllerConfiguration for these, following the same pattern used by BlockchainReader for example

src/Storage/ClaimController.ts
lautarodragan: Yep, we want to avoid that pattern. If you define the functions as arrow functions you don't need to use bind. 

Regarding the `.then`'s, I prefer to avoid using promises directly completely, always using async/await. 

But if you prefer to use `.then` to chain calls, you can remove the `async` from the function declaration since it's not needed.

src/Storage/Service.ts
```diff
@@ -7,11 +7,14 @@ import { childWithFileName } from 'Helpers/Logging'
 import { ClaimController } from './ClaimController'
 import { ServiceConfiguration } from './ServiceConfiguration'
 
+const toMinutes = (milliseconds: number) => 1000 * 60 * milliseconds
```
lautarodragan: Avoid adding such general purpose functions in files dedicated to domain specific things. 

Add them to /Helpers if you think they could be reused, or declare them inside the function that's using them if it's only for readability.

src/Storage/IPFS.ts
```diff
@@ -18,7 +18,7 @@ export class IPFS {
 
   cat = async (hash: string): Promise<string> => {
     const response = await fetch(`${this.url}/api/v0/cat?arg=${hash}`, {
-      timeout: 1000,
+      timeout: 10000,
```
lautarodragan: Awesome catch! Can you make this configurable?

src/Storage/ClaimController.ts
kennylavender: Yeah, I was trying to follow current coding style while also creating a composable flow that avoided a bunch of try/catch everywhere. I was thinking about removing the async from the functions and doing arrow functions but didn't want to break the coding style that much.

src/Storage/IPFS.ts
lautarodragan: Let's go with 10 since you found that worked well.

src/Storage/ClaimController.ts
```diff
@@ -53,51 +54,142 @@ export class ClaimController {
     const logger = this.logger.child({ method: 'download' })
 
     logger.trace({ ipfsHashes }, 'Downloading Claims')
-
-    await this.collection.insertMany(ipfsHashes.map(ipfsHash => ({ ipfsHash, claimId: null })), { ordered: false })
+    await this.collection.insertMany(
```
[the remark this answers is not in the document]
lautarodragan: Not sure either, but I see you merged master into your branch recently, so it should be fine probably.

lautarodragan:
Uh, forgot two small things:
- Please make functions/etc `private` unless you want to explicitly expose them
- Try to order the functions inside a class in the order they call each other. I know it's not always possible since it's not necessarily linear, but it helps read the code.

lautarodragan:
Just noticed some errors logged by the Node in Travis:


```
[2018-05-10T20:15:35.663Z] ERROR (5082 on travis-job-952f7363-c196-44e5-9cf0-0a659f7b5597): Error while Downloading Next Hash
    module: "Storage"
    file: "Service"
    method: "downloadNextHash"
    error: {
      "stack": [
        "TypeError: Cannot read property '_id' of null",
        "    at ClaimController.updateEntryAttempts (/home/travis/build/poetapp/node/dist/babel/src/Storage/ClaimController.js:136:24)",
        "    at <anonymous>",
        "    at process._tickCallback (internal/process/next_tick.js:182:7)"
      ],
      "message": "Cannot read property '_id' of null",
      "type": "TypeError"
    }
```

It's line 136 in the .js source is line 172 in the .ts source.

src/Storage/ClaimController.ts
kennylavender: Not really, Only in the rare case we discussed, where multiple nodes grab the same entry at the same time. So we can probably rename it so its singular.

Do we currently have any sort of naming scheme for timestamps in the database?

Here is a couple off the top of my head that I have seen used before

`lastDownloadAttemptAt` and `downloadSuccessAt`

`lastDownloadAttemptTime` and `downloadSuccessTime`

`timeOfLastDownloadAttempt` and `timeOfDownloadSuccess`

kennylavender:
> Just noticed some errors logged by the Node in Travis:

I am using errors to break out of the composition, I am now logging the error better with a custom throw message.

> Please make functions/etc private unless you want to explicitly expose them

Cool, forgot this, its done now.

> Try to order the functions inside a class in the order they call each other. I know it's not always possible since it's not necessarily linear, but it helps read the code.

Order should be better now ;D


I think I got everthing you requested above except the `lastDownloadSuccess` question which I commented on above.

kennylavender:
I forgot to fix the promise chain with the method bound. Will work on that Monday

kennylavender:
note to self: update the configuration binding to send the configuration needed instead of the entire config object.
```
this.container.bind<ServiceConfiguration>('ServiceConfiguration').toConstantValue(this.configuration)
```

kennylavender:
@lautarodragan should be good to review again

kennylavender:
I tried to use [ramda pipeP](http://ramdajs.com/docs/#pipeP) but the `@types/ramda` package didn't contain the types for `pipeP`. I did find [npm-ramda](https://github.com/types/npm-ramda) that says it contains the types for `pipeP` but I didn't like adding it as a dependency so much, so I created our own asyncPipe function in helpers.

PR #98, opened 2018-05-09
Node docker

WesleyCharlesBlake:
Initial PR to containerize poetapp/node


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+COPY . /usr/src/app/
+RUN chmod a+x /usr/src/app/wait-for-it.sh
+
+WORKDIR /usr/src/app
+
+RUN npm install
+RUN npm run build
+
+CMD [ "npm", "start" ]
```
warrenv: If we can rely on caching layers, rebuilds of the container will go faster using this ordering:
```
FROM node:9.11.1

RUN apt-get update && apt-get install -y rsync

RUN mkdir -p /usr/src/app
WORKDIR /usr/src/app

COPY package.json package-lock.json .
RUN npm install

COPY . /usr/src/app/
RUN chmod a+x /usr/src/app/wait-for-it.sh
RUN npm run build

CMD [ "npm", "start" ]
```

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:9.3.0
```
warrenv: All app repos have been updated to node 9.11.1 now.

Dockerfile
WesleyCharlesBlake: so multiple COPY commands adds an additional layer. each directive in the Dockerfile is seen as Docker image layer on file system. so if package.json* is changed, then Docker will use all other layers that remain untouched, and only create the new layer were necessary.

`COPY . /usr/src/app/` is required to copy everything need (and excludes resources based on [.dockerignore](.dockerignore). So we are copying packer.json* here already.

Including an explicit copy for package.json* will mean two cached layers (at least) will be recreated, instead of just the once cached layer.

Also this OSS, and Docker image is being built by Githubs build servers and pushing to the poetapp/node on docker hub.

src/Configuration.ts
```diff
@@ -32,10 +32,10 @@ export interface LoggingConfiguration {
 }
 
 const defaultConfiguration: Configuration = {
-  rabbitmqUrl: 'amqp://localhost',
-  mongodbUrl: 'mongodb://localhost:27017/poet',
-  ipfsUrl: 'http://localhost:5001',
-  insightUrl: 'https://test-insight.bitpay.com/api',
+  rabbitmqUrl: process.env.RABBITMQ_URL || 'amqp://localhost',
```
lautarodragan: The default configuration is a static immutable object that should never change.

What you want is to add these to the actually loaded configuration, which is then merged with the default configuration.

Since your introducing a new way to load configuration (env vars instead of the conf file), I'll need a bit of refactor. Let me know if you need help working on this.

README.md
```diff
@@ -11,9 +12,11 @@ It's built on top of the Bitcoin's blockchain and [IPFS](https://ipfs.io/).
 
 - [How to Run the Po.et Node](#how-to-run-the-poet-node)
     - [Install](#install)
+    - [Docker Compose](#docker-compose)
+    - [Makefile](#Makefile)
```
lautarodragan: This link doesn't work, at least in the preview. I think you need to change it to lower case.

WesleyCharlesBlake:
Thanks will fix the markdown link.

Will need some help with the immutable stuff. Because a static env file is
going against 12 Factor in the sense that we can't change config in the
environment, which is a big deal in how we would deploy containers to
multiple environments. But think we should discuss it more! Thanks for the
feedback!

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+WORKDIR /usr/src/app
+
+COPY package*.json ./
+RUN npm install
+
+COPY . /usr/src/app/
+
+RUN npm run build
+
+CMD [ "npm", "start" ]
```
kaosmonk: IMHO we should separate the npm build and install processes into a multi-stage Dockerfile and have a lean runtime docker image with only the the bare essentials that our service needs at runtime.
We could even reconsider going with having `--only=prod` and `--only=dev` deps installed in different images so that the runtime only has prod ones in the end.

lautarodragan: In the future, we can make some small modifications to the node so you can `npm i -g poet-node`. Packages published to NPM are already "built" so you don't need to `npm run build`, and when installing globally (`-g`), it's made available as a script (you can then run `poet-node` from any location) and has all its dependencies. 
That way we could have the release docker image just do `npm i -g poet-node` and skip all building etc.

lautarodragan: Actually, it doesn't have to be "in the future", it's an small change which we could address right now if you think it makes sense.

kaosmonk: If we could have `poet-node` as a npm package that one would just install and run the service, that would be awesome and it would help us do the `frost-api` docker image in a more cleaner and sane way.
So if it requires a small change and not much effort to make it happen, I'd say let's do it.

warrenv: How does installing a package that's published on npm help with testing code that has not been published yet? My assumption is that this docker container is being built to run tests and deploy internally. All of that needs to happen before publishing to npm. Is that correct?

kaosmonk: It would definitely help with building pipelines for testing some of our code that depends and relies upon `poet-node` (eg `frost-api`) since we're running the same set of commands, installing and building `node`.
I do not mind building either `poet-node` docker image or a simple package as an artifact that I could use for other purposes as well. So here we should focus ourselves on setting up the pipelines for testing and building `poet-node` artifact, whatever that might be.

lautarodragan: See https://github.com/poetapp/node/pull/103.

warrenv: agreed. It feels like we are tackling too many steps at once. Let's get the pipeline for testing and building the package in place first. That way we can all play around with it and make sure we've built it correctly before building on top of it.

lautarodragan: Agreed. I think this should be a separate discussion — this PR is dockerizing Node. In this scope and contexts, it makes more sense to have the container grab the local sources and build them.

Other projects that consume/interact with the node can use the published package.

lautarodragan:
@WesleyCharlesBlake LGTM. (untested). LMK if you consider it ready to be merged.

kaosmonk:
What about https://github.com/poetapp/node/pull/98#discussion_r187953445? Should we go with multistage Dockerfile?

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    environment:
+      - RABBITMQ_URL=amqp://poet-rabbit
+      - MONGODB_URL=mongodb://poet-mongo:27017/poet
+      - IPFS_URL=http://poet-ipfs:5001
+    hostname: poet-node
+    links:
+      - poet-mongo
+      - poet-rabbit
+      - poet-ipfs
+    command: >
```
kaosmonk: Why do we need a `command` in here when there's one in a Dockerfile?

docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    ports:
+      - "8080:8080"
+      - "4001:4001"
+      - "5001:5001"
+    volumes:
+      - .ipfs-docker:/data/ipfs
+  poet-node:
+    build: .
+    ports:
+      - "18080:18080"
```
kaosmonk: Are these ports correct? I think I saw integration tests are looking for `:1080`? @wzalazar @lautarodragan

WesleyCharlesBlake:
I agree with what has been suggested. Thanks for the input everyone.
However I think we are getting lost in nuances right now. Right now we need
a Docker binary of node for the kubernetes infra. Whilst Murli stage builds
are awesome. I believe it's something we can refactor for in the future,
considering the material benefits it's going to yield right now. Optimizing
builds etc is something we can (and must) address, however not using multi
stage builds is not technical debt that we are introducing. What's more
important is to have a working Dockerfile that we can use for deployment.

Secondly the command section of the docker-compose file is necessary, since
it abstracts from the Dockerfile command to allow the sleep parameter
needed for node to wait for rabbitmq to be ready. It's a simple fix for
something that can be resolved at later stage

WesleyCharlesBlake:
Ports are correct for node

kaosmonk:
RabbitMQ fix is ok.

Re ports, I know documentation says that's the correct port, but I saw integration tests are hitting 1080 as well.

Will double check on my end as well and sync it with @wzalazar

WesleyCharlesBlake:
Can always add more ports to docker-compose.yml if needed

kaosmonk:
Yeah, that's not a problem. It's more like I have different information on ports. :) Will check with @wzalazar but this seems legit.

docker-compose.yml
lautarodragan: poet-node runs on 18080 by default, and consumes mongodb, ipfs and rabbitmq in their default ports.

PR #103, opened 2018-05-14
Allow installing and running poet-node

lautarodragan:
This PR introduces the changes required to be able to publish the node to NPM and make it runnable as an application.

WIth the following commands, one can install and run the Po.et Node:

```sh
npm i -g @po.et/node
poet-node
```

This branch was already [published to NPM](https://www.npmjs.com/package/@po.et/node) for testing purposes, which means you can run the commands displayed above to test it.

We'll need to set up a proper publishing flow using [Travis NPM Deploys](https://docs.travis-ci.com/user/deployment/npm/) soon, so versions can be published by just tagging a commit in GitHub. In the future, we can probably automate GH tagging with a tool that scans commits to `master` for version changes in `package.json`.


package.json
```diff
@@ -1,7 +1,10 @@
 {
-  "name": "poet-node",
-  "version": "1.0.0",
+  "name": "@po.et/node",
+  "version": "1.0.1",
```
warrenv: How are we keeping this up-to-date? If I go to npm, the latest version is 1.0.3, not 1.0.1.

lautarodragan: Sorry! Forgot to include the last commit. Please check now

PR #105, opened 2018-05-15
Fix Confusing Error Logging

### LINKED ISSUE #104: Confusing Error Logging
https://github.com/poetapp/node/commit/ab111295a670b3e5624001e9051da7f938ec6262 introduced a risky code change: the try/catch block now flags caught errors as `INFO`.

This try catch was originally there for possibly unexpected errors thrown down the line not to wind up lost inside a setInterval.

https://github.com/poetapp/node/blob/ab111295a670b3e5624001e9051da7f938ec6262/src/Storage/Service.ts#L34-L49

This change was introduced due to how the new download-retry procedure needs to break out of the pipe when `findEntryToDownload` finds no next entry — since they have all been handled already.

https://github.com/poetapp/node/blob/ab111295a670b3e5624001e9051da7f938ec6262/src/Storage/ClaimController.ts#L91-L138

We need to revert the change in Storage/Service.ts so possible errors get flagged as actual errors, and introduce a different mechanism in `downloadNextHash` to handle this case so the expected exception doesn't bubble up to the service.

kennylavender:
resolves #104 

Created a function to generate an error object with a type. The type can then be used to check if we just want to log info or log an error.


src/Storage/ClaimController.ts
```diff
@@ -79,13 +80,22 @@ export class ClaimController {
...
+        this.findEntryToDownload,
+        this.updateEntryAttempts,
+        this.downloadEntryClaim,
+        this.updateEntryPairs,
+        this.publishEntryDownload
+      )({ retryDelay, maxAttempts })
+      this.logger.info(result, 'Successfully downloaded entry')
+      return result
+    } catch (error) {
+      if (error instanceof NoMoreEntriesException) return this.logger.info(error.message)
```
lautarodragan: I think we can avoid logging this one altogether, since it's part of the normal flow of things. What to you think?

kennylavender: Yup I agree

PR #107, opened 2018-05-18
Docker build tag

WesleyCharlesBlake:
I pulled out the automated docker build, as it was not handling git tag's to push docker tags suffciently. Moved this into travis.

Summary:
1. all Docker images will be tagged as per the git commit , ie to be used in staging/beta stages (not prod eg `poetapp/node:v-e779393-beta`
1. if branch is master, then the Docker image will be tagged latest ( eg `poetapp/node:latest` )
1. when a tag is pushed and released, the image will follow the tag (eg `poetapp/node:v0.0.x` )
    1. If the current build is for a git tag,`TRAVIS_TAG`  variable is set to the tag’s name, regardless of what branch the commit was tagged, this is expected
1. when merging to master, we need to ensure we tag and push the next version of the release. In master, this means that there will be two images with the same Docker hash ( `poetapp/node:latest` and `poetapp/node:v0.0.x`) which is needed, so users can run `docker run poetapp/node ....` and get the latest tag if it doesnt exist, or specify a particular version `docker run poetapp/node:v0.0x` ) to ensure that a pinned version is running/deployed


lautarodragan:
WebStorm and Sublime FTW :)

Other than spacing issues, LGTM.

WesleyCharlesBlake:
@lautarodragan @kaosmonk : I have resolved the indentation , I figured it out: had something to do with the travis gem changing formatting when I used it to encrypt or add fields. Very strange but none the less its resolved. Needs a review, and then can be merged again

PR #120, opened 2018-05-30
Update IPFS version (#88)

### LINKED ISSUE #88: Update IPFS Docker Image
Our make file is using [jbenet/go-ipfs](https://hub.docker.com/r/jbenet/go-ipfs/). There seems to be an official [ipfs/go-ipfs](https://hub.docker.com/r/ipfs/go-ipfs/) one now.

geoffturk:
The latest version of IPFS (v0.4.15) has resolved the memory leak issue. My test instance has flattened out at 72%-75% memory usage for the last 24 hours.

Resolves #88


kaosmonk:
Quick question: this was tested and achieved by just letting the IPFS running without any load/traffic/usage? I wonder if it would be any different with load/traffic and if we would be able to see any changes and deviations from the figures you reported in terms of resource usage.

kaosmonk:
What I am trying to say is that usually one can find memory hogs and leakage only when the service is set with load and high traffic for example. And services tend to behave differently when under load ofc.

geoffturk:
During the 3.5 days the server has been running the node, IPFS has accumulated 123mb of data on the server. Here's how the memory usage looked over that period of time:

image

Another server with the same spec (2CPU, 4gb RAM) and running node (but with ipfs-v0413) and similar data usage just ran away all of the memory:

image

Even after restarting the server, it ran OOM again in less than 24 hours.

WesleyCharlesBlake:
Sorry somehow missed my input here! LGTM, I will delete my PR!

PR #134, opened 2018-06-05
Update koa-body to version 4.0.0 🚀

lautarodragan:
https://github.com/poetapp/node/pull/133


warrenv:
@lautarodragan Yes, there are still seems to be some problem with this repo in particular and the greenkeeper-lockfile package.

warrenv:
I will use this PR to dig deeper into the problem.

PR #136, opened 2018-06-05
Updated isClaim to isValidClaim 

### LINKED ISSUE #31: Possible Exploit: downloadClaim
Storage's `downloadClaim` currently throws in three cases:
- if ipfs.cat throws (connection issue or other reason)
- if the contents of the file aren't valid JSON
- if the contents of the file are valid JSON but not a valid Claim

In the first case the download should just be retried in the future. Ensuring the different modules stay in sync and are fail-safe is a complex issue, not covered by this task.

In the second and third cases, it's possible for an attacker to broadcast transactions that are valid Po.et transactions, pointing to a real IPFS file, but storing non-json or a json object that doesn't pass the isClaim test in it.

Such an attack is simple to execute and would effectively crash all Po.et Nodes in the network, or even worse, execute arbitrary code.

This attack should be covered by an integration test.

jwicks31:
In Storage's `downloadClaim`

Resolves #31


jwicks31:
There are unit tests that check isValidClaim only returns true if the following pass:
- correct claim keys
- correct claim signature
- correct claim id
- correct claim publicKey

Does that cover the testing scenario or would you still like an integration test added?

warrenv:
Just unit tests are ok for now.

PR #145, opened 2018-06-12
IPFS failure detail

### LINKED ISSUE #121: IPFS Download Failure Detail
Related to #53 

Our procedure to download files from IPFS currently takes any failure at all and marks the download request as failed, increasing the attempt count for it, but ignoring the details of what exactly went wrong.

https://github.com/poetapp/node/blob/3e624760b0e7d546898b966cb393161893debd02/src/Storage/ClaimController.ts#L76-L99

https://github.com/poetapp/node/blob/3e624760b0e7d546898b966cb393161893debd02/src/Storage/ClaimController.ts#L180-L190

We need to add support to distinguish between network timeout errors and "not a claim" errors.

### LINKED ISSUE #144: Download Claim Issue, poet-js isValidClaim
Possible issue with the nodes usage of `isValidClaim` from poet-js, it was introduced in #136

I keep getting this error during the downloading of claims:

```
[2018-06-12T05:00:49.268Z] ERROR: claim.dateCreated.getTime is not a function
```

It seems to be from calling the method `getTime` on a string instead of a date object.

I checked the tests in poet-js and `isValidClaim` expects a claims `dateCreated` property to be a date object, not a string.

This is blocking #20, #142

lautarodragan:
Fixes #121 
Fixes #144 

## Description of Changes

This PR introduces finer-grained information logging of error types in the Storage module. In particular, it introduces the `failureReason` and `failureType` properties in the database, and a modification to the ipfs download retry system, not to retry `HARD` failures.

#### `failureReason`

This field is informative. Can be any of the following: `INVALID_JSON`, `INVALID_CLAIM`, `IPFSGeneric`, `IPFSTimeout`. These values are defined in the string enum `FailureReason`.

#### `failureType`

This field affects the retry system. If can be `SOFT` or `HARD`. `HARD` failure types won't be retried.

### Function Modifications

All modified functions are in the Storage module.

#### `ClaimController.findEntryToDownload`

This function's query to mongodb was now adapted to ignore ipfsHashes marked with a `HARD` failure.

#### `ClaimController.download`

This function has been fixed. It will now ignore `Duplicate key` MongoDB errors.

#### `ClaimController.downloadClaim` 

This function was removed. Its code was adapted into `downloadEntryClaim`, which previously merely acted as a wrapper for `downloadClaim`. 

`downloadClaim` was only being called from `downloadEntryClaim`.

#### `Router.onPoetTimestampsDownloaded`

This function now properly catches all errors thrown by the call to `claimController.download`, logging them at `error` level.

Previously, it would let any error bubble up, which would wind up as an unhandled promise rejection, which are deprecated and will terminate the process in new versions of NodeJS.

### New Make Commands

Four new make commands were introduced to help manual debugging of Storage issues: `ipfs-unpin-all` and `ipfs-gc`, which delete all files in the IPFS container, `reset-storage`, which removes all information of all entries of the storage collection but the ipfsHash itself and `print-storage-errors`, which prints the count by `failureReason`.

### New Fetch Error Helper and node-fetch typings

Got all this data from reading [node-fetch's source code](https://github.com/bitinn/node-fetch).

I'll try and make a couple of PRs to node-fetch and @types/node-fetch this Friday or the next, but we'll need to keep this code in our code base for now.

Example run of `print-storage-errors`:
```
$ sudo make print-storage-errors 
docker exec -i poet-mongo mongo poet < ./queries/storage-errors.mongo
MongoDB shell version v3.4.10
connecting to: mongodb://127.0.0.1:27017/poet
MongoDB server version: 3.4.10
{ "_id" : "INVALID_JSON", "count" : 339 }
{ "_id" : "IPFSTimeout", "count" : 1400 }
{ "_id" : null, "count" : 8551 }
bye
```

## Test

```
> db.blockchainReader.find({}, {blockHeight: 1, _id:0}).sort({blockHeight: 1}).limit(1)
{ "blockHeight" : 1282538 }
> db.blockchainReader.find({}, {blockHeight: 1, _id:0}).sort({blockHeight: -1}).limit(1)
{ "blockHeight" : 1325267 }
> db.storage.count()
10750
> db.storage.find({claimId: {$ne: null}}).count()
8672
> db.storage.find({claimId: null, failureReason: null}).count()
0
> db.storage.find({claimId: null, failureReason: {$ne: null}}).count()
2078
> db.storage.aggregate([   { $match: {claimId: null} } ,   { $group: {_id: "$failureReason", count: { $sum: 1 } } } ])
{ "_id" : "IPFS_TIMEOUT", "count" : 1 }
{ "_id" : "INVALID_JSON", "count" : 2077 }
> db.storage.find({claimId: null, failureReason: 'IPFS_TIMEOUT'}, {_id: 0, ipfsHash: 1})
{ "ipfsHash" : "QmeYHwNLoCQUQ82FzFRsJ9vRK8GSiFEBW3ycwTV8x7GYFS" }
```

### Configuration used

```
{
    "minimumBlockHeight": 1282537,
    "downloadIntervalInSeconds": 0.1,
    "downloadRetryDelayInMinutes": 1,
    "downloadTimeoutInSeconds": 30,
}
```

## TODO

- The error-checking logic in `downloadEntryClaim.download` could be moved to the IPFS client
- `downloadNextHash.handleErrors` could be made into an async pipe of middlewares
- `download` is checking the error code against the [magic number 11000](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err#L272).

### PR Review Checklist


lautarodragan:
Thanks @geoffturk and @warrenv !

I've pushed more changes and updated the test results.

Out of 10750 ipfsHashes 2077 failed due to invalid JSON and 1 due to timeout.

The one that times out doesn't exist in production either, so it's probably some test someone run in their local machine.

2077 broken IPFS files is worrying, but at least now we have concrete data.

lautarodragan:
Ready for full review :)

lautarodragan:
@poetapp/quality-engineering Travis is failing even though the tests are passing. Do you guys know what's going on?
These are the last lines of the travis output:
```
The command "npm run test" exited with 0.
after_script
0.19s$ greenkeeper-lockfile-upload
'ipfs-add-failure-detail' is not a Greenkeeper branch
Done. Your build exited with 1.
/home/travis/.travis/job_stages: line 156:  5114 Terminated              npm start
```

warrenv:
I see lint failures at
https://travis-ci.org/poetapp/node/builds/391607806?utm_source=github_status&utm_medium=notification

lautarodragan:
Thanks @warrenv, hand't seen that. Will fix

lautarodragan:
Now passing 🚀

src/Helpers/FetchError.ts
```diff
@@ -0,0 +1,19 @@
...
+  NoRedirect = 'no-redirect',
+  MaxRedirect = 'max-redirect',
+  UnsupportedRedirect = 'unsupported-redirect',
+  MaxSize = 'max-size',
+  InvalidJson = 'invalid-json',
+}
+
+const TimeoutFetchErrorTypes: ReadonlyArray<string> = [FetchErrorType.RequestTimeout, FetchErrorType.BodyTimeout]
+
+export const isFetchError = (error: any): error is FetchError => error instanceof Error && error.name === 'FetchError'
```
jwicks31: Can we unit test this?

src/Helpers/FetchError.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const isFetchTimeoutError = (error: Error) => isFetchError(error) && TimeoutFetchErrorTypes.includes(error.type)
```
jwicks31: Can we unit test this?

src/Storage/ClaimController.ts
```diff
@@ -79,23 +87,44 @@ export class ClaimController {
...
-      )({ retryDelay, maxAttempts })
-      this.logger.info('Successfully downloaded entry')
-      return result
-    } catch (error) {
-      if (error instanceof NoMoreEntriesException) return this.logger.trace(error.message)
-      this.logger.error(error)
+  } = {}): Promise<void> {
+    const logger = this.logger.child({ method: 'downloadNextHash' })
+
+    const updateEntryFailureReason = (ipfsHash: string, failureType: FailureType, failureReason: FailureReason) =>
```
jwicks31: This is being used with await below. Should this be an async function?

src/Storage/ClaimController.ts
```diff
@@ -178,10 +210,37 @@ export class ClaimController {
   }
 
   private downloadEntryClaim = async ({ entry, ...rest }: { entry: Entry }) => {
+    const { ipfsHash } = entry
+    const downloadClaim = async () => {
```
jwicks31: Should downloadClaim take ipfsHash as an argument rather than grabbing from lexical scope?

src/Storage/ClaimController.ts
```diff
@@ -178,10 +210,37 @@ export class ClaimController {
...
+      }
+    }
     const logger = this.logger.child({ method: 'downloadEntryClaim' })
-    logger.trace('starting claim download')
-    const claim = await this.downloadClaim(entry.ipfsHash)
-    logger.trace('finished claim download', claim)
+
+    logger.trace({ ipfsHash }, 'Starting claim download')
+
+    const serialized = await downloadClaim()
```
jwicks31: We should probably pass in the hash to downloadClaim.

kennylavender:
@jwicks31 and I reviewed above.

src/Storage/ClaimController.ts
lautarodragan: It's not necessary for a function that returns a promise to be defined as async. 

I was letting the type of the function be inferred since it's a one liner that runs a query, but if you guys think it's confusing maybe declaring it as `: Promise<whatever mongodb returns>` could do?

src/Helpers/FetchError.ts
lautarodragan: Yeah! Good catch. I'll add that test.

src/Helpers/FetchError.ts
lautarodragan: Yeah! Good catch. I'll add that test.

src/Storage/ClaimController.ts
lautarodragan: I intentionally avoided doing that since this function only exists inside the parent function `downloadEntryClaim`, and `ipfsHash` is an argument of `downloadEntryClaim`. I thought it was just adding noise.

But still I'd like your opinion on this. If you find it more readable and less error prone to explicitly pass the argument, we can try that.

src/Storage/ClaimController.ts
lautarodragan: See comment above.

lautarodragan:
Thanks @jwicks31 and @kennylavender for the review! I've left a couple of responses to your comments. And I'll start working on adding the two unit tests you guys pointed out.

lautarodragan:
Alright, new info: regarding the unit tests, I've discussed with Warren and the best approach for now is to skip them since he'll be adding riteway very soon. I'll create a ticket to track them, and add the unit tests using the riteway framework asap, but in a separate PR.

src/Storage/ClaimController.ts
jwicks31: Would it allow us to move it outside of the parent and unit test it if we pass it as an argument?

lautarodragan: @jwicks31 it would require mocking. Which is making me realize I could change the way this is coded so it could be unit tested. I'm thinking something along this lines:

Instead of having

```ts
    const downloadClaim = async (ipfsHash: string) => {
      try {
        return await this.ipfs.cat(ipfsHash)
      } catch (error) {
        if (isFetchTimeoutError(error)) throw new IPFSTimeoutError(ipfsHash)
        else throw new IPFSGenericError(ipfsHash, error)
      }
    }
```

We could have something like 

```ts
const errorToIPFSError = (error, ipfsHash) => isFetchTimeoutError(error)
  ? new IPFSTimeoutError(ipfsHash) 
  : new IPFSGenericError(ipfsHash, error)

const downloadClaim = (ipfsHash: string) =>
  this.ipfs.cat(ipfsHash).catch(errorToIPFSError)
```

In this case we could unit-test `errorToIPFSError`.

lautarodragan: Alright, got it. I've introduced a new function, Promise.prototype.rethrow, which acts like Promise.catch but throws the result of calling the callback. 

`rethrow` allows us to declare pure error-translating functions like this:

```ts
export const errorToIPFSError = (ipfsHash: string) => (error: Error) =>
  isFetchTimeoutError(error)
    ? new IPFSTimeoutError(ipfsHash)
    : new IPFSGenericError(ipfsHash, error)
```

This function can then be used like this:

```ts
    const downloadClaim = async (ipfsHash: string) =>
      this.ipfs.cat(ipfsHash).rethrow(errorToIPFSError(ipfsHash))
```

I'll update the PR's description with info on Promise.prototype.rethrow.

`errorToIPFSError` is a 100% pure function, no side effects, no throws. It can be easily unit tested.

`downloadClaim` doesn't need to be tested at all, since it's a plain call to two functions which themselves can be tested individually.

lautarodragan:
@jwicks31 @kennylavender I've now added a README.md file with documentation about the Storage module, including stuff about the failure type and reason.

Adding a README.md file right inside the Module's folder has the advantage of being right at hand when working on the module's code. Another nice perk is that it'll be displayed when navigating into the folder in GitHub.

jwicks31:
@lautarodragan awesome! It was very thorough! Cool to have it display as you enter the folder in Github as well!

src/Storage/ClaimController.ts
kennylavender: Yup your right, don't think I noticed the implicit return while looking at Jesse's screen :D

I think we could remove the `await` from the function calls though since we don't need the response?

lautarodragan: Not sure which one you’re talking about but in general we need to await even if we’re not using the result so errors get thrown. On my phone atm so hard to get context.

PR #142, opened 2018-06-11
Implement Claim Batching

kennylavender:
resolves #20 
resolves #158 
resolves #159
resolves #151 

Instead of timestamping individual claims as they come into the node, we want to group them and timestamp a batch of claims only once every n minutes.

## Overview

This PR introduces 2 new modules, BatchWriter and BatchReader and a few modifications to handle routing changes between the other modules. Both of these modules include READMEs with details on files and responsibilities.

### Writing Requirements
- [x] provide configuration and acceptable defaults
- [x] empty directories should not be timestamped
- [x] duplicate directory hashes should not be timestamped
- [x] BatchWriter entries should be marked complete by a successTime after being successfully timestamped
- [x] do not allow duplicate entries in the collection

### Reading Requirements
- [x] provide configuration and acceptable defaults
- [x] previous hashes on the blockchain should be ignored with the version bump
- [x] should read the directory and successfully download all the file hashes into IPFS
- [x] do not allow duplicate entries in the collection
- [x] each file hash should have an entry in the works collection, and also contain the ipfs directory hash

#### `src/Configuration.ts`
Bumped the node version up. This allows us to read only directory timestamps from the blockchain and ignore timestamps from previous versions. This simplifies the logic required to sync the nodes because we now simply ignore hashes that were generated in previous versions and were not batched.

Added the following configuration items
- batchIntervalInSeconds: 600, every 10 minutes the BatchWriter will create a batch
- readDirectoryIntervalInSeconds: 30, every 30 seconds the BatchReader will read a directory and publish the files hashes of the directory

#### `src/BlockchainWriter/Router.ts`
Adjusted the router to consume `BatchWriter` messages. The `BatchWriter` now batches file hashes into a directory hash and publishes the directory hash to be timestamped.

#### `src/Messaging/Messages.ts`
Added more message types. Message types are global at the moment, but it might be nice if each module in the future provides its own messages.

The messages I added are similar to messages used in async redux actions, there is a `request`, `success` and `failure` for every message. This should give us some more flexibility in the future because we can now subscribe to success and failure messages from other modules if we want to build some logic around those messages.

#### `src/Storage/Router.ts`
Adjusted the router to consume `BatchReader` messages. The `BatchReader` now reads directories that come from the `BlockchainReader` and then publishes the file hashes so that they can be downloaded by the `Storage` module.

#### `src/index.js`
Added and started the 2 new services, __BatchReader__, and __BatchWriter__.

### Directions

- Setup the configuration like below
- Start the node
[...]

PR #166, opened 2018-06-20
Fix BlockchainWriter collection update

kennylavender:
Found this issue while working on #142.

When the same hash been supplied to `requestTimestamp` multiple times there will be multiple entries in the database. The previous version of this function would find and update the first entry that matched the hash even if it already had a txId. Which means that the duplicates will never get updated and are constantly in the running to be timestamped. This pr fixes the update function to update a collection entry with no txId.

### PR Review Checklist


lautarodragan:
Nice catch. So for some reason we have duplicated entries in the BlockchainWriter DB collection — `requestTimestamp` is being called more than once with the same argument.

I don't think we want to timestamp each of these duplicated entries, though. A single IPFS hash shouldn't exist in many transactions. That would be a huge increase of overall network usage and transaction fees with no real advantage.

I think it probably makes more sense for the BlockchainWriter collection to have a unique index over the `ipfsHash`, same as the Storage one, and for `requestTimestamp` to silently ignore `Duplicated Index` errors, so timestamp requests for the same IPFS hash aren't introduced more than once.

https://github.com/poetapp/node/blob/e75ba844c520b09a4532ee55a1e98ea374d19d58/src/Storage/ClaimController.ts#L72-L86

https://github.com/poetapp/node/blob/e75ba844c520b09a4532ee55a1e98ea374d19d58/src/Storage/Storage.ts#L74-L78

What do you think?

kennylavender:
Nice, yes that sounds better to me! Its what I tried to do originally but was not sure how to make things unique in mongo collection. I will close, create a new issue.

PR #170, opened 2018-06-21
Implement Claim Batching - directoryCollection module proposal

kennylavender:
part of #142 

Proposing the following file change to the directory collection module and the related changes to make it possible.

Here were my goals

#### No database
Remove the need for the directory collection module to know about the database because it only needs to know about the single collection.

#### Receive a collection, don't define one.
The directory module does not need to know about the specific collection name if it just receives a collection as an argument. This allows the higher level modules ( Router.js or Storage.js ) to define what the collection names are and which modules manipulate those collections. It also means we could reuse a collection module because we can create multiple instances and give them different collections. We might be able to create a collection module that handles multiple use cases since lots of our hash/retry/attempt logic is looking very similar.

### Responsibility
The directory collection module should handle all logic related to the collection, I think setting the unique indexes should be done from this module, not in other modules. This makes logic related to the unique indexes testable with integration tests of this single module. 

Handling the responsibility in the directory collection module
https://github.com/poetapp/node/blob/claim-batching-proposal-01/src/Storage/DirectoryCollection.ts#L37

Vs handling it externally
https://github.com/poetapp/node/blob/claim-batching-proposal-01/src/Storage/Storage.ts#L80-L84

### Typescript Noise
remove typescript typing noise, try to separate types from functions, still working on this, not sure how I like the current code. Being new to typescript please point out anything you see that could be improved here for sure.

### Easier Testing, less mocking
Completing the above items should make the directory collection module much easier to test with integration tests if we decide to do so.

@lautarodragan please take a look, let me know if you like, don't like these changes and suggestions, thanks!

If you like these changes we can apply them to the other modules in the future.


lautarodragan:
Very nice!

Double checking to make sure I understand: It's basically introducing a new layer to the application, sort of like the DAO pattern. Is this correct?

If so, I avoided it originally but seeing how the Storage/ClaimController file is looking right now, I think it's a good introduction.

Agree 100% with your first three points: "No Database", "Receive a Collection, Don't Define One" and "Responsibility". 

### Typescript Noise

I'm not a big fan of this style. Reminds me of .h files in c++. I think type information is too valuable to be separated from the implementation. Eric seems to favour it, though. 

Considering you're providing defaults for all functions, maybe we can let the type inference do its magic and avoid them? 

I'm not really sure, though. Providing explicit type declarations can be redundant when type inference is doing its job, but it's also a safety net saving us from accidentally changing a default value and introducing a bug. 

And ReadonlyArrays can't be inferred.

### Using Closures instead of Classes

I don't really see how this change is needed. Can't the same be achieved by changing `constructor(@inject('DB') db: Db) {` to `constructor(@inject('DirectoryMongoCollection') collection: Collection) {`?

The nice thing about DI is that you don't need to use if you don't want to. The DI framework scans the annotations and instantiates classes with the correct parameters if you use it, but the class is still a class with a normal constructor in the end. This allows you to instantiate the class as you would normally do from your unit tests: `const dirCol = new DirectoryCollection(myMockedCollection)`.

## Word of caution

Whenever I say module, I refer to one of the 5 modules the Po.et node is made of - each of which function like separate applications with no dependencies on each other. There's never a clear definition of what "module" means without context, but I'd rather avoid calling other things modules to avoid confusion. In the case of Router.js or Storage.js, I usually just call them files or layers.

kennylavender:
> Double checking to make sure I understand: It's basically introducing a new layer to the application, sort of like the DAO pattern. Is this correct?Had not heard this pattern name before, but I looked it up quickly, looks like it!

kennylavender:
> Double checking to make sure I understand: It's basically introducing a new layer to the application, sort of like the DAO pattern. Is this correct?

Had not heard this pattern name before, but I looked it up quickly, I think this might be more of a facade pattern than but I could be wrong.

> If so, I avoided it originally but seeing how the Storage/ClaimController file is looking right now, I think it's a good introduction.

Cool! exactly :D

### Using Closures instead of Classes

Passing the collection into the class is not the reason why I switched to a factory function. The reason I switched is becaue I couldn't figure out a good way to await the unique index setup with a class. Maybe you have an idea? 

`await collection.createIndex({ hash: 1 }, { unique: true })`
https://github.com/poetapp/node/blob/claim-batching-proposal-01/src/Storage/DirectoryCollection.ts#L37
( Realized i provided a broken link above, I fixed it )

### Typescript Noise
Yeah, I was thinking we could just use the defaults and drop all the types, but then I realized we need all the function types to build up the returned interface `DirectoryCollection` that is used by `Storage.js` and `Router.js`

I almost like the idea of moving these types out of the file and into a file located right next to it `DirectoryCollection.d.ts` is that a possiblity? I tried to do it last night, but after a few hours of messsing with typescript stuff I settled for whats in this pr. Let me know if you have any ideas on this.

kennylavender:
> Whenever I say module, I refer to one of the 5 modules the Po.et node is made of - each of which function like separate applications with no dependencies on each other. There's never a clear definition of what "module" means without context, but I'd rather avoid calling other things modules to avoid confusion. In the case of Router.js or Storage.js, I usually just call them files or layers.

Thanks, I caught on to this for the most part ;) I have been flip/flopping on this personally I like the idea of calling the individual files modules since that is how node refers to them, and then calling the groupings of modules 'features' or something else, but modules also makes sense there as well ¯\_(ツ)_/¯ idk. This sort of brings up the idea that maybe we could define something as a team and document it?

lautarodragan:
> Thanks, I caught on to this for the most part ;) I have been flip/flopping on this personally I like the idea of calling the individual files modules since that is how node refers to them, and then calling the groupings of modules 'features' or something else, but modules also makes sense there as well ¯_(ツ)_/¯ idk. This sort of brings up the idea that maybe we could define something as a team and document it?

Yeah I'm thinking the same thing, we should document it. I've been thinking about moving them to a new folder, `src/modules/...`, so it's clear they are modules and they are not the same as `Helpers`, `Messaging` and `Extensions`.

And you're right - hadn't thought of it before, but it's true that to NodeJS those are modules, so it's even more confusing. Can't come up with a proper name though. In reality they are separate applications, but calling them so would be confusing since we're treating them as one, too.

kennylavender:
Ah, to continue using a class we can require that an async init method be called after instantiation. The async init method will set up the unique index.

So instead of
```
this.directoryCollection = await createDirectoryCollection({
  collection: this.dbConnection.collection('storageDirectories'),
})
```
We would need to create the instance then call and await the init method.
```
this.directoryCollection = new DirectoryCollection({
  collection: this.dbConnection.collection('storageDirectories'),
})
await this.directoryCollection.init()
```

If you prefer that better I can update it to do that.

kennylavender:
> Yeah I'm thinking the same thing, we should document it. 
:+1: 

> Can't come up with a proper name though. In reality they are separate applications, but calling them so would be confusing since we're treating them as one, too.

Yup, that exactly the problem I couldn't find words for, nicely done :D

lautarodragan:
> If you prefer that better I can update it to do that.

That looks better to me, personally. 

I've grown to like class constructors that don't do anything other than storing the constructor's arguments in read-only class variables (for fun, see Scala's Case Classes — they are beautiful](https://docs.scala-lang.org/tour/case-classes.html). Above all, I expect class constructors to never have side effects. That way I can rest assured nothing bad can happen from a class instantiation — explicitly calling the `initialize` is where dragons lie.

> Yup, that exactly the problem, I couldn't find words to describe it, nicely done :D

Thanks :smile: Still not sure how to tackle this, though. I'll create an issue and tag you and others for feedback.

> I almost like the idea of moving these types out of the file and into a file located right next to it DirectoryCollection.d.ts is that a possibility? I tried to do it last night, but after a few hours of messing with typescript stuff, I settled for whats in this pr. Let me know if you have any ideas on this.

I tried that in the past and failed, too. I'm not sure you can use `.d.ts` files _inside_ your project, for your own codebase. I think declaration files exists only to type external code bases.

Also: I REALLY don't like that approach :disappointed: In c++ I always wound up switching between files to make sure I was doing the right thing. Having type definitions right at hands helps me code the function. 

I think it's similar to TDD: write the test first, see it fail. Then write the function's contract - what it expects, what it should return. If the function and arguments are named and typed correctly, the function's body should follow intuitively.

But maybe it's just a matter of personal preference.

ericelliott:
I think we need a broader discussion of our use of classes and decorators in general.

Without having much familiarity with the codebase, I get the following impressions:

1. Everything is too tightly coupled and requires too much mocking. See ["Mocking is a Code Smell"](https://medium.com/javascript-scene/mocking-is-a-code-smell-944a70c90a6a) for a rather in-depth discussion on this topic.
2. It appears that we have some implicit dependencies on code that's introduced via decorator syntax. That is an anti-pattern. Implicit dependencies are an anti-pattern regardless of the cause. I wrote about them in more depth in an article on [functional mixins](https://medium.com/javascript-scene/functional-mixins-composing-software-ffb66d5e731c#415f). Implicit dependencies are a common problem with mixins and decorators. Essentially, they provide an illusion of modularity, but the reality of a monolithic design.

BTW, these comments are not familiarity bias. I am very familiar with these OO patterns because I spent most of my early career specializing in OO design.

In general, there are only two good reasons to use the `class` keyword in JavaScript:

1. Performance optimization (only when you need to access millions of props per second in a tight loop, e.g., render loops. React and RxJS use it for this purpose, for example. You should *very rarely* need to do it in actual application code), or,
2. Framework hooks, e.g., React, Angular

There are LOTS of reasons for these guidelines. You can read more about them:
* ["What's the Difference Between Class and Prototypal Inheritance?](https://medium.com/javascript-scene/master-the-javascript-interview-what-s-the-difference-between-class-prototypal-inheritance-e4cd0a7562e9)
* ["Factory Functions vs Constructor Functions vs Classes"](https://medium.com/javascript-scene/javascript-factory-functions-vs-constructor-functions-vs-classes-2f22ceddf33e)
* ["How to Use Classes and Sleep at Night"](https://medium.com/@dan_abramov/how-to-use-classes-and-sleep-at-night-9af8de78ccb4)

I'm available to answer questions/objections on Slack. =)

kennylavender:
> Above all, I expect class constructors to never have side effects. That way I can rest assured nothing bad can happen from a class instantiation — explicitly calling the initialize is where dragons lie.
This is a good point, probably could/should have made my factory function return an init function to handle the initial side effects as well.

>  for fun, see Scala's Case Classes — they are beautiful.
Took a look, nice, these seem much closer in behavior/format to JS Factory Functions than ES6 classes.

@ericelliott @lautarodragan, let me know once you have synced on this.

lautarodragan:
@kennylavender we'll sync on this on Monday and maybe continue during the week. 

Until then, I think the sanest thing to do is going ahead with using classes in this PR to avoid blocking you and avoid introducing two parallel styles in the code base, and handle a migration in a separate PR if we decide to do so.

I've been guilt of introducing style changes gradually in the past, or being in favor of other team members doing so, and that never ended well. 

If we do decide to go ahead with this it should be properly planned.

PR #183, opened 2018-06-26
Prevent duplicate DB entries in BlockchainWriter

### LINKED ISSUE #167: Prevent duplicate DB entries in BlockchainWriter
To avoid timestamping duplicate hashes we need to remove the possibility of adding duplicate hashes to the BlockchainWriter's DB collection.

See examples here on how to create a unique index for ipfsHash.

https://github.com/poetapp/node/blob/e75ba844c520b09a4532ee55a1e98ea374d19d58/src/Storage/ClaimController.ts#L72-L86

https://github.com/poetapp/node/blob/e75ba844c520b09a4532ee55a1e98ea374d19d58/src/Storage/Storage.ts#L74-L78

Original discussion #166

kennylavender:
resolves #167 

### Description of Changes

There is no need to timestamp the same hash more than once, it would be a waste of money. This PR prevents duplicate entries from being added to the blockchains database queue by making ipfsHash a unique index.

It follows the same structure the Storage module implemented to prevent duplicate entries because it requires the least amount of code changes.

### PR Review Checklist


kennylavender:
@lautarodragan ready for review

PR #202, opened 2018-07-02
Validation incoming data

### LINKED ISSUE #172: Validate incoming data
With the goal to protect the API, we have to implement a validation system. I suggest creating a middleware with Joi.

wzalazar:
fix #172 

### Description of Changes

- Creates middleware validation incoming data
- Unit test of middleware incoming data
- Integration test
- Adds Joi and types
- Validates query and params


wzalazar:
@warrenv @lautarodragan ready to review 👍

src/API/Middlewares/RequestValidationMiddleware.ts
```diff
@@ -0,0 +1,27 @@
...
+  try {
+    const { query, body, params, options } = requestValidationParams
+    /* tslint:disable:no-unused-expression */
+    body && (await Joi.validate(context.request.body, body(context.request.body), options))
+    query && (await Joi.validate(context.query, query(context.query), options))
+    params && (await Joi.validate(context.params, params(context.params), options))
+    return next()
+  } catch (exception) {
+    if (context.logger)
+      context.logger.error({ exception }, 'Error Caught at Middleware - Will Return Internal Server Error')
```
lautarodragan: Does `Joi.validate` throw if the validation fails?

If so, this isn't an error and we shouldn't log it as such. Don't know whether we care about logging it at all, even. Maybe log it as `debug`?

If we decide to keep this log, the logging message should be updated too. It's not an Internal Server Error, and it's not an Error. I'm thinking `RequestValidationMiddleware rejected a request.`

test/Integration/GetWorksByPublicKey.ts
```diff
@@ -122,4 +122,12 @@ export class GetWorksByPublicKey {
       Expect(claims[i].dateCreated.toISOString()).toBe(expectedClaims[i].dateCreated.toISOString())
     }
   }
+
+  @AsyncTest()
+  @TestCase('')
```
lautarodragan: I like this test!

What do you think of adding other test cases that should fail, such as passing something that isn't a string maybe?

src/API/Middlewares/RequestValidationMiddleware.ts
```diff
@@ -0,0 +1,27 @@
+import { IllegalArgumentException } from '@po.et/poet-js'
+import * as Joi from 'joi'
```
lautarodragan: Can you add an empty line after the imports?

src/API/Middlewares/RequestValidationMiddleware.ts
```diff
@@ -0,0 +1,27 @@
...
+interface RequestValidationParams {
+  readonly query?: (values: object) => {}
```
lautarodragan: If `values` isn't an array, rename to `value`. I'm also thinking something like `querySchema`, `bodySchema` and `paramsSchema` would be more descriptive.

src/API/Middlewares/RequestValidationMiddleware.ts
```diff
@@ -0,0 +1,27 @@
...
+    /* tslint:disable:no-unused-expression */
+    body && (await Joi.validate(context.request.body, body(context.request.body), options))
+    query && (await Joi.validate(context.query, query(context.query), options))
+    params && (await Joi.validate(context.params, params(context.params), options))
+    return next()
+  } catch (exception) {
+    if (context.logger)
+      context.logger.error({ exception }, 'Error Caught at Middleware - Will Return Internal Server Error')
+
+    throw new IllegalArgumentException(exception)
```
lautarodragan: IllegalArgumentException takes an string as a parameter, not an object. If we're intentionally passing the exception object and expecting `.toString()` to be called on it, let's do that here to avoid room for interpretation.

We could also improve the `IllegalArgumentException` to support more precise parameters if needed.

test/Integration/GetWorksByPublicKey.ts
```diff
@@ -122,4 +122,12 @@ export class GetWorksByPublicKey {
...
+  async getWorks422(publicKey: string) {
```
lautarodragan: Can you rename this test to something like `getWorksShouldFailWith422WhenPassingAnInvalidArgument`?

src/API/Middlewares/RequestValidationMiddleware.ts
wzalazar: it's an object and match with `ctx.query`, `ctx.params` and `ctx.request.body`.

What do you think if we change the name to `queryRequestValues`, for example?

src/API/Middlewares/RequestValidationMiddleware.ts
wzalazar: with the error message is enough. I can send the `exception .message`. 

Do you agree?

test/Integration/GetWorksByPublicKey.ts
wzalazar: Yeah, I like it!

src/API/Middlewares/RequestValidationMiddleware.ts
wzalazar: done

src/API/Middlewares/RequestValidationMiddleware.ts
wzalazar: Does Joi.validate throw if the validation fails?

yes, `Joi.validate ` throw an error

What do you think, it's important to log it?

src/API/Middlewares/RequestValidationMiddleware.ts
lautarodragan: I meant renaming `values` (the parameter of the function), not the keys of `RequestValidationParams`. Do you want to rename the keys, too?

src/API/Middlewares/RequestValidationMiddleware.ts
lautarodragan: Yeah, that sounds good.

src/API/Middlewares/RequestValidationMiddleware.ts
lautarodragan: I'd go with leaving it there out of caution, but just under `debug` so we don't get a flood of error messages that trigger alarms.

src/API/Middlewares/RequestValidationMiddleware.ts
wzalazar: for example:

```
interface RequestValidationParams {
  readonly query?: (queryRequestValues: object) => {}
  readonly body?: (bodyRequestValues: object) => {}
  readonly params?: (paramsRequestValues: object) => {}
  readonly options?: object
}
```

src/API/Middlewares/RequestValidationMiddleware.ts
wzalazar: ok

test/Integration/GetWorksByPublicKey.ts
wzalazar: in the cases, when we send query or params values, we only have the option to pass `string` or `empty` values, because body parser will convert to a string. For this reason, I only check empty option.

Do you have any idea?

src/API/Middlewares/RequestValidationMiddleware.ts
```diff
@@ -0,0 +1,28 @@
+import { IllegalArgumentException } from '@po.et/poet-js'
+import * as Joi from 'joi'
+
+interface RequestValidationParams {
+  readonly query?: (queryRequestValues: object) => {}
```
wzalazar: @lautarodragan do you like the names now?

tests/unit/index.ts
```diff
@@ -1,3 +1,4 @@
 // All unit test files should be imported here. For example,
 // import '../../src/foo.test'
 import '../../src/canary.test'
```
wzalazar: I can remove this test canary and the comment. What do you think?

src/API/Router.ts
```diff
@@ -39,8 +41,16 @@ export class Router {
     this.configuration = configuration
     this.workController = workController
 
-    this.koaRouter.get('/works/:id', this.getWork)
-    this.koaRouter.get('/works', this.getWorks)
+    const getWorksSchema = {
```
lautarodragan: Just thought: what if we move the `params` and `query` that we're defining inline in the call to `RequestValidationMiddleware` to here?

So it would look something like this:

```ts
const getWorksSchema = {
  params: {
    id: Joi.string().required(),
  }
}
...
this.koaRouter.get('/works/:id', RequestValidationMiddleware(getWorksSchema), this.getWork)
```

This way, the `getWorksSchema` is actually the complete schema for that endpoint. Otherwise, we're really defining part of the schema here only, and the last bit inline in the call to `RequestValidationMiddleware`.

Well, that's my reasoning, anyway. 

Let me know what you think.

test/Integration/GetWorksByPublicKey.ts
lautarodragan: Makes sense. After all, all query and body params are strings really, and then we convert them to whatever.

src/API/Router.ts
wzalazar: yes, you can define both ways. I think it's more simple like as you say.

PR #215, opened 2018-07-11
Implement Helpers/IPFSClient.ts

### LINKED ISSUE #187: Implement Helpers/IPFSClient.ts
part of #186 

Implement the helper.

Combine the IPFS required methods from

- [x] src/Storage/IPFS.ts
- [x] src/BatchReader/IPFS.ts in #142 
- [x] src/BatchWriter/IPFS.ts in #142

kennylavender:
resolves #187 

Simply implements the Helpers/IPFSClient.ts module, hooking it up and using it in other modules will be done in future pull requests.

- no dependency injection
- created as a class for consistency
- implements relevant IPFS methods needed for
  - Storage
  - BatchReader
  - BatchWriter

### PR Review Checklist

### DIFF
--- a/src/Helpers/IPFSClient.ts
+++ b/src/Helpers/IPFSClient.ts
@@ -1,0 +1,103 @@
+import * as FormData from 'form-data'
+import fetch from 'node-fetch'
+import * as str from 'string-to-stream'
+
+import { secondsToMiliseconds } from 'Helpers/Time'
+
+interface IPFSConfiguration {
+  readonly ipfsUrl: string
+}
+
+interface FetchConfiguration {
+  readonly timeout: number
+}
+
+enum Type {
+  File = 'File',
+  Directory = 'Directory',
+}
+
+interface IPFSObject {
+  readonly Name?: string
+  readonly Hash: string
+  readonly Size?: number
+  readonly Type?: Type
+  readonly Links?: ReadonlyArray<IPFSObject>
+}
+
+interface LSResult {
+  Arguments: {
+    [key: string]: string
+  }
+  Objects: {
+    [key: string]: IPFSObject
+  }
+}
+
+
+type addFileToDirectory = (directoryHash: string, fileName: string, filehash: string) => Promise<IPFSObject>
+
+type addText = (text: string, x?: FetchConfiguration) => Promise<IPFSObject>
+
+type cat = (s: string, x?: FetchConfiguration) => Promise<string>
+
+type createEmptyDirectory = () => Promise<IPFSObject>
+
+type getDirectoryFileHashes = (s: string) => Promise<ReadonlyArray<string>>
+
+type ls = (s: string) => Promise<LSResult>
+
+
+export class IPFS {
+  private readonly url: string
+
+  constructor(configuration: IPFSConfiguration) {
+    this.url = configuration.ipfsUrl
+  }
+
+  addFileToDirectory: addFileToDirectory = async (directoryHash, fileName, fileHash) => {
+    const response = await fetch(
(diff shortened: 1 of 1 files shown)


src/Helpers/IPFSClient.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  addFileToDirectory: addFileToDirectory = async (directoryHash, fileName, fileHash) => {
+    const response = await fetch(
+      `${this.url}/api/v0/object/patch/add-link?arg=${directoryHash}&arg=${fileName}&arg=${fileHash}`
+    )
+    return await response.json()
+  }
+
+  addText: addText = async (text: string, fetchConfiguration) => {
+    const formData = new FormData() // { maxDataSize: 20971520 }
```
kennylavender: Does converting the text to form data belong in this method? or is it the callers responsibility?

src/Helpers/IPFSClient.ts
```diff
@@ -0,0 +1,98 @@
...
+      method: 'post',
+      body: formData,
+    })
+
+    return await response.json()
+  }
+
+  cat: cat = async (hash, fetchOptions) => {
+    const response = await fetch(`${this.url}/api/v0/cat?arg=${hash}`, fetchOptions)
+    return response.text()
```
kennylavender: Why does cat return `response.text()` instead of `response.json()`?

src/Helpers/IPFSClient.ts
lautarodragan: I think having a layer that abstracts this away makes sense. In my mind, the IPFSClient should hide away any details of _how_ to communicate with the API.

src/Helpers/IPFSClient.ts
kennylavender: Cool, so I think we should be able to change it now to `response.json()` without breaking anything I think.

kennylavender: Changed `cat` type to return Response["buffer"] from node-fetch as you suggested, ts then marked the `cat` function red, I then changed the cat function to `response.json()` and it removed the red, so it seems to like the change.

src/Helpers/IPFSClient.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  addFileToDirectory: addFileToDirectory = async (directoryHash, fileName, fileHash) => {
+    const response = await fetch(
+      `${this.url}/api/v0/object/patch/add-link?arg=${directoryHash}&arg=${fileName}&arg=${fileHash}`
+    )
+    return await response.json()
+  }
+
+  addText: addText = async (text: string, fetchOptions) => {
```
warrenv: Is the type 'string' needed here given the definition on line 37?
btw, I am finding the use of 'type' makes the functions easier to read.

kennylavender: Nope, good catch thanks! :D

Same here, I like that I can read through all the types and interfaces without the implementation details.

lautarodragan:
@kennylavender this PR needs [Update @po.et/tslint-rules to the latest version 🚀 #233](https://github.com/poetapp/node/pull/233), right?

kennylavender:
@lautarodragan Yes :D

lautarodragan:
Merged :smile:

lautarodragan:
Restarting Travis, PR passed but push failed with weird output...

lautarodragan:
@kennylavender is this one ready to be merged?

kennylavender:
i think the ipfs library you found would probably be the better option, we could close this?

PR #216, opened 2018-07-11
Claim batching proposals 2

lautarodragan:
- Call fileDAO.start and directoryDAO.start
- mv src/Error.ts to src/Extensions/Error.ts (completely unrelated to Claim Batching)
- alter BatchWriter flow a bit
- BlockchainWriter no longer publishes a BlockchainWriterTimestampRequestCreated message.
- BlockchainWriter doesn't care about ipfsFileHashes — only about the ipfsDirectoryHash
- BatchWriter.ClaimController.createNextBatch now returns the ipfsDirectoryHash only, and takes care of calling `completeHashes`. The reasoning behind this is that the BatchWriter module doesn't really care what other modules may do with the batch it created. From its scope, the batch has been created and that's it.
- BatchWriter.ClaimController.createNextBatch now checks whether there are actually any items to batch. If not, it'll throw, preventing the creation of an empty directory that will not be used.
- BatchWriter.Router.onBlockchainWriterTimestampRequestCreated no longer exists, since that message no longer exists either. It's not needed, since `createNextBatch` takes care of calling `completeHashes` now.

BUG found


src/View/Router.ts
```diff
@@ -34,6 +34,7 @@ export class Router {
       Exchange.BatchReaderReadNextDirectorySuccess,
       this.onBatchReaderReadNextDirectorySuccess
     )
+    await this.messaging.consume(Exchange.BatchWriterCreateNextBatchSuccess, this.onBatchWriterCreateNextBatchSuccess)
```
lautarodragan: @kennylavender we were missing adding the ipfsDirectoryHash to works when the Claim Batch is being created in this node rather than being downloaded from others

src/BlockchainWriter/Router.ts
```diff
@@ -34,29 +34,23 @@ export class Router {
     const logger = this.logger.child({ method: 'onBlockchainWriterRequestTimestampRequest' })
 
     const messageContent = message.content.toString()
-    const { ipfsFileHashes, ipfsDirectoryHash } = JSON.parse(messageContent)
+    const { ipfsDirectoryHash } = JSON.parse(messageContent)
```
lautarodragan: @kennylavender The BlockchainWriter doesn't really need to know anything about individual claims now, just the batch

src/BatchWriter/Router.ts
```diff
@@ -26,22 +27,20 @@ export class Router {
   async start() {
     await this.messaging.consume(Exchange.ClaimIPFSHash, this.onClaimIPFSHash)
     await this.messaging.consume(Exchange.BatchWriterCreateNextBatchRequest, this.onBatchWriterCreateNextBatchRequest)
-    await this.messaging.consume(
```
lautarodragan: @kennylavender In my mind, the BatchWriter doesn't need to know anything about other systems. Its responsibility is to create batches and announce them, but ends there. It's up to other systems which want to do something with that Claim Batch to implement their own logic.

src/View/WorkController.ts
```diff
@@ -45,7 +45,7 @@ export class WorkController {
     await Promise.all(
       poetTimestamps.map(timestamp =>
         this.collection.updateOne(
-          { 'timestamp.ipfsHash': timestamp.ipfsHash },
+          { 'timestamp.ipfsDirectoryHash': timestamp.ipfsHash },
```
lautarodragan: @kennylavender we have a bug here. With Claim Batching this function no longer works as expected. It should be updating all entries in the works collection with information on the timestamp. It's an upsert, so if we had no info on the claim yet, it would create an empty entry that only had the timestamp and later add claim info to it.

It now no longer works because it's creating a single entry for N claims. The timestamp also has a field named incorrectly now: `timestamp.ipfsHash` should now be `timestamp.ipfsDirectoryHash` I believe.

Since there's no 1-to-1 mapping any more, we'll need to figure out some logic to handle this correctly.

kennylavender:
Nice, catches, i've merged a couple of the simple things

src/View/WorkController.ts
kennylavender: I see, I thought I already checked for this but must have missed it. I think if it listens for the BatchReader instead of BlockchainReader it should solve it.

src/BatchReader/Service.ts
```diff
@@ -32,6 +37,16 @@ export class Service {
   }
 
   private readNextDirectory = async () => {
-    this.messaging.publish(Exchange.BatchReaderReadNextDirectoryRequest, '')
+    const logError = (logger: Pino.Logger) => (error: any) =>
```
kennylavender: try/catch and await might look a bit nicer for this change

src/View/Router.ts
kennylavender: I did that on purpose because after a transaction is created and added to a block, the block would then be read and we would get the ipfsDirectoryHash that way. I thought about doing both but wasn't sure if it was redundant. I think its fine to have both ways so Il merge this too.

src/BatchWriter/Router.ts
lautarodragan: Yeah. Your system made sense. I'm just thinking we could reduce overall complexity by giving BatchWriter limited responsibility.

The pro of your system is that in case of failure the directory that had been generated would be discarded, I think, and a new one created on the next cycle. That way we'd avoid queuing more and more timestamp requests if they fail.

My reasoning in that case was that we may want them to start piling up: if we just keep adding claims to the pending-batch queue, that's could be a sign of something going wrong or it could just be increased usage. But if timestamp requests start piling up, it's 100% guaranteed to be an issue and we can easily spot it and act on it.

src/View/WorkController.ts
lautarodragan: Yup. It's not the easiest, either. I was thinking one way to handle this would be to add a new collection to view/api: something like `timestamps`. We store all the timestamp information there, and when new claims are added to the `works` collection, timestamp metadata is looked up in the `timestamps` collection and copied over. (We'd have two collections but API would still query one, without joins).

Copying data over from one collection to the other would require careful examination of the flows of the code. My gut feeling is it would be super easy to leave an undetected bug that generates invalid data over time.

src/View/Router.ts
lautarodragan: Yep, we want to have both just so we don't have to wait 10 minutes to see that data in the API.

In the future we'll be able to add block confirmations to ensure the data is consistent. For now, if there are discrepancies between the data read from a block and the one written on the creation process we can either log it or create a ticket about it so we don't forget.

src/BatchReader/Service.ts
lautarodragan: Sure! Let's go with that :)

PR #219, opened 2018-07-12
Updates GetWorks to Get All Works

jwicks31:
### Description of Changes

- Adds getWorks to WorkController
- Updates Router to handle getWorks

### PR Review Checklist

### DIFF
--- a/src/API/Router.ts
+++ b/src/API/Router.ts
@@ -48,6 +48,10 @@ export class Router {
     }
 
     const getWorkSchema = {
+      query: {},
+    }
+
+    const getWorkByPublicKeySchema = {
       query: {
         publicKey: Joi.string().required(),
       },
@@ -55,6 +59,7 @@ export class Router {
 
     this.koaRouter.get('/works/:id', RequestValidationMiddleware(getWorksSchema), this.getWork)
     this.koaRouter.get('/works', RequestValidationMiddleware(getWorkSchema), this.getWorks)
+    this.koaRouter.get('/works?publicKey=', RequestValidationMiddleware(getWorkByPublicKeySchema), this.getWorksByPublicKey)
     this.koaRouter.post('/works', this.postWork)
 
     this.koa.use(helmet(SecurityHeaders))
@@ -84,12 +89,19 @@ export class Router {
   private getWorks = async (context: KoaRouter.IRouterContext, next: () => Promise<any>) => {
     this.logger.trace({ query: context.query }, 'GET /works')
 
+    const works = await this.workController.getWorks()
+
+    context.body = works
+  }
+
+  private getWorksByPublicKey = async (context: KoaRouter.IRouterContext, next: () => Promise<any>) => {
+    this.logger.trace({ query: context.query }, '/works?publicKey=')
     const publicKey = context.query.publicKey
     const works = await this.workController.getByPublicKey(publicKey)
 
     context.body = works
   }
-
+  
   private postWork = async (context: KoaRouter.IRouterContext, next: () => Promise<any>) => {
     this.logger.trace({ body: context.request.body }, 'POST /works')
 
--- a/src/API/WorkController.ts
+++ b/src/API/WorkController.ts
@@ -26,6 +26,11 @@ export class WorkController {
     return this.collection.findOne({ id }, { fields: { _id: false } })
   }
 
+  async getWorks(): Promise<any> {
+    this.logger.trace({ method: 'getWorks'}, 'Getting Works from DB')
+    return this.collection.find({}, { fields: { _id: false } }).toArray()
+  }
+
   async getByPublicKey(publicKey: string): Promise<any> {
     this.logger.trace({ method: 'getByPublicKey', publicKey }, 'Getting Works by Public Key from DB')
     return this.collection.find({ publicKey }, { fields: { _id: false } }).toArray()


src/API/Router.ts
```diff
@@ -48,13 +48,22 @@ export class Router {
...
+
+    const getWorkByPublicKeySchema = {
       query: {
         publicKey: Joi.string().required(),
       },
     }
 
     this.koaRouter.get('/works/:id', RequestValidationMiddleware(getWorksSchema), this.getWork)
     this.koaRouter.get('/works', RequestValidationMiddleware(getWorkSchema), this.getWorks)
+    this.koaRouter.get(
```
warrenv: This is a problematic way of defining routes. What if other query params were valid for /works? How would you ensure the user puts 'publicKey' as the first param?

This logic should be part of the /works path as it is a means of filtering works. The filtering logic should be inside that route instead of making a new one. 

Further reading:
https://medium.com/@moschan/when-should-you-use-path-variable-and-query-parameter-a346790e8a6d

warrenv:
We really should put a feature toggle system in place before we start adding new behavior like this. @kennylavender @lautarodragan @geoffturk @wzalazar

src/API/Router.ts
```diff
@@ -91,16 +82,8 @@ export class Router {
...
-
-    const works = await this.workController.getWorks()
-
-    context.body = works
-  }
-
-  private getWorksByPublicKey = async (context: KoaRouter.IRouterContext, next: () => Promise<any>) => {
     this.logger.trace({ query: context.query }, '/works?publicKey=')
-    const publicKey = context.query.publicKey
+    const publicKey = context.query.publicKey ? context.query.publicKey : undefined
```
warrenv: I don't think this variable is necessary?  context.query.publickKey will either exist or not so it can be passed directly in the next line. There is the question of what happens if there's no context.query?

Using TDD will help with finding these kinds of edge cases as the code is developed.

kennylavender:
@warrenv I am not sure feature toggles are necessary in this case because it's just a slight modification to getWork handler but we def need to get feature toggles setup soon.

src/API/WorkController.ts
```diff
@@ -26,9 +26,11 @@ export class WorkController {
     return this.collection.findOne({ id }, { fields: { _id: false } })
   }
 
-  async getByPublicKey(publicKey: string): Promise<any> {
+  async getByPublicKey(publicKey?: string): Promise<any> {
```
lautarodragan: `getByPublicKey` no longer represents what this method does. Maybe `getByFilters`?

src/API/WorkController.ts
```diff
@@ -26,9 +26,11 @@ export class WorkController {
...
     this.logger.trace({ method: 'getByPublicKey', publicKey }, 'Getting Works by Public Key from DB')
-    return this.collection.find({ publicKey }, { fields: { _id: false } }).toArray()
+    return publicKey
+      ? this.collection.find({ publicKey }, { fields: { _id: false } }).toArray()
```
lautarodragan: This isn't going to scale. As we add another param we'll have to add all possible combinations.

Maybe we can do something like 

```ts
interface WorksFilters {
  readonly publicKey?: string
}

async getByFilters(worksFilters: WorksFilters = {}): Promise<any> {
  const definedFilters = Object.entries(worksFilters)
    .filter(([key, value]) => value !== undefined) 
    .toObject()
  return this.collection.find(definedFilters, { fields: { _id: false } }).toArray()
}
```

We can use `undefined` as a special case since MongoDB doesn't allow comparing against `undefined` anyway.

```
> db.works.find({"noField": undefined})
Error: error: {
	"ok" : 0,
	"errmsg" : "cannot compare to undefined",
	"code" : 2,
	"codeName" : "BadValue"
}
```

src/API/Router.ts
```diff
@@ -82,8 +82,7 @@ export class Router {
   }
 
   private getWorks = async (context: KoaRouter.IRouterContext, next: () => Promise<any>) => {
-    this.logger.trace({ query: context.query }, 'GET /works')
-
+    this.logger.trace({ query: context.query }, '/works?publicKey=')
     const publicKey = context.query.publicKey
     const works = await this.workController.getByPublicKey(publicKey)
```
lautarodragan: If we implement the  `getByFilters` as I mention in my other comment and if Joi validates that no unknown query parameters are being passed, we could just `const works = await this.workController.getByPublicKey(context.query)` here.

@wzalazar do you know if Joi rejects unexpected parameters?

wzalazar: Joi reject all request that contains unknown parameter unless you has enabled the option to force accept it parameters

lautarodragan: Thanks @wzalazar! Sounds like we can `const works = await this.workController.getByPublicKey(context.query)` :smile:

src/API/WorkController.ts
jwicks31: Ok I made these changes. I don't think I'm understanding why we would need to check for undefined though. Isn't Joi only allowing us to pass in known parameters that aren't empty? 
```js
async getByFilters(worksFilters: WorksFilters = {}): Promise<any> {
  return this.collection.find(worksFilters, { fields: { _id: false } }).toArray()
}
```
`Array.toObject` I may have implemented differently than you were expecting as well. Let me know if I need to rework it.

src/API/WorkController.ts
```diff
@@ -3,10 +3,10 @@ import { inject, injectable } from 'inversify'
 import { Collection, Db } from 'mongodb'
 import * as Pino from 'pino'
 
+import 'Extensions/Array'
```
lautarodragan: Add imports with side effects to the root of the application whenever possible:

https://github.com/poetapp/node/blob/e75ba844c520b09a4532ee55a1e98ea374d19d58/src/index.ts#L1-L10

src/API/WorkController.ts
lautarodragan: I'm not sure either now. Dunno what I was thinking originally...

test/Integration/GetWorks.ts
```diff
@@ -0,0 +1,96 @@
...
+    const allElementsAreClaims = !claims.find(
+      (claim: Claim) => !isClaim(claim)
+    );
+
+    Expect(allElementsAreClaims).toBeTruthy();
+  }
+
+  @AsyncTest()
+  @TestCase()
+  async getWorksShouldReturnExpectedFields(
```
jwicks31: This test fails after I let the node run for long enough and I'm assuming it is because there are some claims that aren't valid being posted?

test/Integration/Helper.ts
```diff
@@ -16,9 +16,18 @@ export class Client {
     return fetch(this.url + '/works/' + id)
   }
 
+  getWorks() {
+    return fetch(this.url + '/works/')
+  }
+
+  getWorksTest(test?: string) {
```
jwicks31: Not really sure this is necessary. But included it just in case.

test/Integration/GetWorks.ts
```diff
@@ -0,0 +1,96 @@
+/* tslint:disable:no-relative-imports */
```
jwicks31: I wasn't sure that I could test for a certain number of claims since it is a changing number.

lautarodragan: I was hard-coding an expected amount of works in some cases because the database is being prepared for tests. See [test/PrepareDB.ts](https://github.com/poetapp/node/blob/master/test/PrepareDB.ts).

I don't know what the best approach is since we don't want individual tests to depend on the order they run or other tests running first, but I guess we can at least rely on the database being prepared... so you can assume the number of works should be equal or greater than the prepared amount.

It would be nice to have a test that checks that the output from `GET /works` differs from `GET /works?publicKey=...`.

test/Integration/Helper.ts
lautarodragan: Looks good!

test/Integration/GetWorks.ts
jwicks31: Ok I've added a test that checks that `GET /works` returns more claims than `GET /works?publicKey=...` and a test to ensure that `GET /works` returns at least 3 claims.

lautarodragan:
@kennylavender / @warrenv wanna review this one again?

src/API/WorkController.ts
```diff
@@ -26,9 +30,11 @@ export class WorkController {
...
   }
 
-  async getByPublicKey(publicKey: string): Promise<any> {
-    this.logger.trace({ method: 'getByPublicKey', publicKey }, 'Getting Works by Public Key from DB')
-    return this.collection.find({ publicKey }, { fields: { _id: false } }).toArray()
+  async getByFilters(worksFilters: WorksFilters = {}): Promise<any> {
+    const definedFilters = Object.entries(worksFilters)
+      .filter(([key, value]) => value !== undefined)
+      .toObject()
+    return this.collection.find(definedFilters, { fields: { _id: false } }).toArray()
```
kennylavender: Are there any security issues or misuses that could arise from passing entire query object into collection.find()? I am thinking it might be better to explicity pass in filters we accept, and maybe a test case should be created to avoid this.

kennylavender: @lautarodragan @warrenv ping, I think @jwicks31 might be awaiting your feedback here.

lautarodragan: Thanks Kenny. 

Joi is currently validating that no unexpected fields arrive in the API layer, so nothing unexpected should reach this function.

Ultimately the controller layer shouldn't depend other layers "behaving nice", so to speak. It should be atomically functional, including any validation it wants to do. 

But for now I'm OK doing it this way — it's not the only place this layer depends on other layers for validation and not an immediate issue. Maybe we could create an issue?

warrenv: I agree with @lautarodragan - it's ok for now but it is a potential crack for sql injection.
I'd like to see a unit test put in place for it in this PR at least.

jwicks31: This PR does have a test to ensure that if we pass a filter that isn't accepted we get the correct response. I think I may be misunderstanding though. Below is the test and the helper function.

```ts
  @AsyncTest()
  @TestCase('test')
  async getWorksShouldFailWith422WhenPassingAnInvalidArgument(test: string) {
    const response = await this.client.getWorksTest(test)

    Expect(response.status).toBe(422)
  }
```
```ts
  getWorksTest(test?: string) {
    return fetch(this.url + '/works/?test=' + test)
  }
```

lautarodragan:
@jwicks31 anything blocking this one?

jwicks31:
Yeah. I realize that @warrenv asked about unit test, not integration test. I'd like to try and break it out and add some.

jwicks31:
I added the tests in #254.

src/Helpers/getDefinedFilters.ts
```diff
@@ -0,0 +1,13 @@
+// tslint:disable-next-line
+import '../Extensions/Array'
```
wzalazar: absolute paths are not working?

src/Helpers/getDefinedFilters.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface Query {
+  readonly publicKey?: string
+}
+
+export const getDefinedFilters = (worksFilters: Query = {}): Query => {
+  const definedFilters = Object.entries(worksFilters)
+    .filter(([key, value]) => value !== undefined)
+    .toObject()
```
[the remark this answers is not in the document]
jwicks31: That makes sense. My thought was that it would eventually be used to define the filters as well. Something like: 

```ts
const filters: ReadonlyArray<string> = ['publicKey']

export const getDefinedFilters = (worksFilters: Query = {}): Query => Object.entries(worksFilters)
    .filter(([key, value]) => value !== undefined && filters.includes(key))
    .toObject()
```

src/Helpers/getDefinedFilters.ts
lautarodragan: I wonder why this didn't get picked up by ts-lint...?

jwicks31: The absolute path wasn't working for some reason. I added the line to ignore the rule, because it wasn't letting me build with the absolute path.

lautarodragan: I'll try to debug it

src/Helpers/getDefinedFilters.ts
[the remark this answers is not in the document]
jwicks31: Updated this in #262. Let me know if you think it is OK.

PR #222, opened 2018-07-12
copy node specific interfaces from poet-js

kennylavender:
### Description of Changes

Copy node specific interfaces and bitcoin code to this repo so we can modify as necessary for claim batching pr. Updated imports to use the newly defined interface file.

### PR Review Checklist


src/Bitcoin.ts
```diff
@@ -0,0 +1,31 @@
+import * as bitcore from 'bitcore-lib'
+
+import { TransactionPoetTimestamp } from 'Interfaces'
+
+export const PREFIX_POET = Buffer.from('POET')
+export const PREFIX_BARD = Buffer.from('BARD')
+
+export function getPoetTimestamp(tx: bitcore.Transaction): TransactionPoetTimestamp {
```
lautarodragan: Move to /Helpers?

kennylavender: I was actually thinking the BatchReader because its only used there? but Helpers could work too, not sure what makes most sense to you?

kennylavender: Moved to helpers as that works good

lautarodragan: Yeah that's probably better.

lautarodragan: Oops, didn't see your last message. "Yeah that's probably better." was meant for "I was actually thinking the BlockchainReader because its only used there? but Helpers could work too, not sure what makes most sense to you?"

Either way is fine though

lautarodragan:
@warrenv those unit tests will require knowledge about Insight API which we're moving away from very soon. Also, these are not new functions, Kenny just copied them over from poet-js, so no new behavior is being introduced. 

Do you still consider we need unit tests for this?

warrenv:
We should recognize the general rule that anytime code is touched, unit
tests should be added if not already present.
This will be the most efficient way to back-fill missing unit tests and get
the coverage metric we are shooting for (80%).

In this case, I'm ok with putting it off until the refactor away from
Insight.

PR #224, opened 2018-07-13
Adds openapi.yaml file for Swagger.io

jwicks31:
### Description of Changes

Adds a swagger openapi.yaml file. 

- To test load file in [swagger](http://editor.swagger.io/#)

### PR Review Checklist


jwicks31:
The only thing that isn't working right now is the example for posting a work. It returns the error `Claim's type must be Work, not string`. I have tried using both 
```yaml
type: 
  Work: "Work"
```
and
```yaml
type: "Work"
```

warrenv:
This is a great start @jwicks31! 
How about replacing the boilerplate in the summary and descriptions with actual text?

openapi.yaml
```diff
@@ -0,0 +1,149 @@
+openapi: 3.0.1
+info:
+  title: Poet Node API
```
lautarodragan: We should always use "Po.et" wherever possible.

openapi.yaml
```diff
@@ -0,0 +1,149 @@
...
+        code:
+          type: string
+        message:
+          type: string
+      required:
+        - code
+        - message
+    Timestamp:
+      type: 'object'
+      properties:
```
lautarodragan: Timestamp has a few other properties. Look at the [timestamp interface declaration](https://github.com/poetapp/poet-js/blob/master/src/Interfaces.ts#L55) for reference, and technical tab of [this work](https://explorer.po.et/works/c481118d5ac0527fca2023d3650744c723bb742fcb5977871df4a3d74d0bc1cb) for an example.

jwicks31: I went ahead and added these in the current file.

PR #261, opened 2018-07-26
Implement Helper/IPFSClient.ts with ipfs-api

### LINKED ISSUE #187: Implement Helpers/IPFSClient.ts

kennylavender:
work in progress

resolves #187 

I've added a simple npm script to the branch to log out results of all the current functions. You will need the ipfs container running. `npm run test:ipfs`

## types
Not quite sure how we will type Buffers and Streams that come from ipfs-api

I am not familiar at all with Buffers and Streams, I think il need to spend some time learning what they are and how to use them.

Also not quite sure if we should type the ipfs-api package or just the IPFSClient? typing the ipfs-api package will require typing things like [DAGNodes](https://github.com/ipld/js-ipld-dag-pb) and DAGLinks

### Buffer
the cat function of ipfs-api returns a Buffer from https://www.npmjs.com/package/buffer, not sure how we can type this. Its supposed to be 100% identical to nodejs's Buffer, so using the global Buffer object as the returned type might work?

### ipfs-api functions return objects with methods
many of the ipfs-api functions return objects like [DAGNodes](https://github.com/ipld/js-ipld-dag-pb) not sure if we will ever need these. These objects need to have .toJSON() called on them to get regular json data from them.

### hash vs multhash
file functions return ipfs objects with a hash property, but object/directory functions return directories with a multihash property and no hash property. This can be hidden by the IPFSClient facade but it will take some figuring out.

### PR Review Checklist


kennylavender:
@lautarodragan do you think this is adding unnecessary complexity and we should just stick with using the api directory or is there going to be some benefits to using the ipfs-api package?

src/Helpers/IPFSClient.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  console.log('emptyDirectory', emptyDirectory)
+
+  const ls = await ipfs.ls('QmcGcrtDGWvij6mmv5Uyqod1ddmycxX1MWKk5VLp8QJTkx')
+  console.log('ls', ls)
+
+  const addFileToDirectory = await ipfs.addFileToDirectory(emptyDirectory.hash, addTextResult.hash, addTextResult.hash)
+  console.log('addFileToDirectory', addFileToDirectory)
+}
+
+run()
```
warrenv: How long should it take for this to run? I seem to be stuck waiting for the 'ls' output after 2-3 minutes.

kennylavender: Ipfs can take a long time to resolve a file/directory, not sure of it's time out. My ipfs node that has that directory is currently offline so it probably won't find it. You can maybe add your own directory to ls

src/Helpers/IPFSClient.test.ts
```diff
@@ -0,0 +1,28 @@
+import { IPFS } from './IPFSClient';
+
+const port = 5001
+const ipfsUrl = `http://localhost:${port}`;
```
warrenv: we'll need a way to override 'port' and 'ipfsUrl' with a env var - currently the test won't run from inside the node container, only outside.

kennylavender: This isn't a real test file, and the values are just hard coded on purpose for the moment so I could expirement with the differences.

do we want integration tests for these types of files? If so more than happy to create the integration tests.

If so I will need an ipfs executable available to start from my test file or the ipfs container running

kennylavender: Here is the original pr that uses the api directly https://github.com/poetapp/node/pull/215

src/Helpers/IPFSClient.test.ts
lautarodragan: Can't we test `ls` with a dynamic value rather than a hard-coded one? Maybe doing the ls test against the `emptyDirectory` before and after calling `addFileToDirectory`?

lautarodragan:
Another note on publishing @po.et/ipfs-client: I've seen at least one or two devs wanting a simple function to create and update immutable IPFS directories, but it's never straight forward or well documented. I had to ask around until someone pointed me in the right direction, and was lucky enough someone had actually come across the same issue and had a solution.

I think it would add value to the community.

kennylavender:
Yup, @po.et/ipfs-client, sounds good to me.

kennylavender:
Close, we are going to create @po.et/ipfs-client. We have also decided not to use the package `ipfs-api`

Issue #95, opened 2018-05-08
Pagination


lautarodragan:
We need to add pagination to the Node. When requesting Works by Public Key it currently returns everything. 

We'll also need this for #94

PR #263, opened 2018-07-27
Pagination

### LINKED ISSUE #236: GetWorks Integration Tests
* [x] Add Integration tests for getWorks (getByFilter with no filter)

jwicks31:
Resolves #236 & #95 

### Description of Changes

- Updated getByPublicKey to getByFilter in WorkController to get all works if no public key is provided
- Updates getWorksSchema to optionally include public key
- Adds offset parameter to get-works query for pagination
- Updates getByFilters to return 10 works at a time

### TODO:
* [x] Integration test
* [x] Fix typing for omitUndefined + unit test

### PR Review Checklist


src/API/WorkController.ts
```diff
@@ -33,8 +34,15 @@ export class WorkController {
 
   async getByFilters(worksFilters: WorksFilters = {}): Promise<any> {
     this.logger.trace({ method: 'getByFilters', worksFilters }, 'Getting Work by Filters from DB')
+    const { offset } = omitUndefined(worksFilters)
+    const skip = parseInt(offset, 10) > 10 ? parseInt(offset, 10) - 10 : 0
+    const limit = parseInt(offset, 10) || 10
```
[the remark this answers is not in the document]
jwicks31: I think I must have been thinking limit was the upper bound of the works being retrieved. Do you think it would be good to just hard code at 10?

jwicks31:
In order to add integration tests for this I would need to add more Claims to the `PrepareDB.ts` & `Claims.ts` correct? Is there a preference on which claims I should use for this?

src/API/WorkController.ts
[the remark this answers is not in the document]
lautarodragan: I'd make it configurable just to avoid a magic number... defaulting to 10.

src/Helpers/GetDefinedFilters.ts
```diff
@@ -0,0 +1,14 @@
+// tslint:disable-next-line
```
lautarodragan: If this file is specific to the API module, we should probably move it there.

src/Helpers/GetDefinedFilters.ts
```diff
@@ -0,0 +1,14 @@
...
+import '../Extensions/Array'
+
+interface Query {
```
lautarodragan: `Query` says more about _how_ we're using this interface than _what_ it is. Maybe simply calling it `WorksFilters`?

src/API/WorkController.ts
```diff
@@ -33,8 +35,17 @@ export class WorkController {
 
   async getByFilters(worksFilters: WorksFilters = {}): Promise<any> {
     this.logger.trace({ method: 'getByFilters', worksFilters }, 'Getting Work by Filters from DB')
+    const { offset }: { offset?: string } = omitUndefined(worksFilters)
+    const skip = parseInt(offset, 10) || 0
     const definedFilters = getDefinedFilters(worksFilters)
-    return this.collection.find(definedFilters, { fields: { _id: false } }).toArray()
+    return this.collection
+      .find(definedFilters, { fields: { _id: false } })
```
lautarodragan: `fields` is deprecated, use `projection` instead. See http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#find

lautarodragan:
> In order to add integration tests for this I would need to add more Claims to the PrepareDB.ts & Claims.ts correct? Is there a preference on which claims I should use for this?

@jwicks31 yep! feel free to add whatever claims you think are needed to PrepareDB. 
Only preference would be using new ones, different from the ones we're already using, to avoid confusion. 

There's a `test/Script.ts` that you can freely modify and `node dist/babel/test/Script.js` to create new claims out of a private key and set of attributes. This file is used nowhere in the project, so there's no risk of breaking anything — it's only there for running it manually.

lautarodragan:
@jwicks31 I just noticed this PR is against the `getWorks` branch — is that ok?

WesleyCharlesBlake:
I've deployed poetapp/node:v-e1aef48-beta (as built from the last [commit](https://github.com/poetapp/node/pull/263/commits/e1aef4809b1ccc48c87c38131cbf4dcf6e5382b8) in this PR) to staging 

http://node.staging.po.et/works 

The liveness and readiness probes are passing now 👍

jwicks31:
@lautarodragan We could delete the `getWorks` branch and merge this one into master instead, but this was a branch off of `getWorks`.

jwicks31:
Commit 269b9dc was only 1 line change that shouldn't have had any effect on the tests. I'm a little lost on why the test is failing now for getWorksShouldReturnClaims.

warrenv:
If you look at the travis output, there's only 1 failing integration test. This happens from time to time and you just need to re-run the travis test. I've kicked it off again for you in this case.

jwicks31:
It doesn't anymore. I just manually checked all of the claims that we are inserting into the test database   and tested that `!claims.find((claim: Claim) => !isClaim(claim))` should return the correct value. They all passed when checking manually.

ericelliott:
> isClaim checks that it is an instanceof Date

All `instanceof` checks are fragile (flat out lie across execution contexts). Is there a better way to do that check?

lautarodragan:
@ericelliott `isClaim` is not introduced in this PR nor defined in the node. [It's defined in poet-js](https://github.com/poetapp/poet-js/blob/master/src/Interfaces.ts#L12).

Improving the implementation of `isClaim` is a pending issue. Do you want to wait to merge this PR until we improve it in poet-js?

If we want a quick fix, I think we could probably change `object.dateCreated instanceof Date` to `!isNaN(new Date(object.dateCreated))`. What do you think?

ericelliott:
@lautarodragan I'll leave those choices to you and @jwicks31 to work out, as long as we're moving forward. I trust you guys. =)

lautarodragan:
For the record: created https://github.com/poetapp/poet-js/issues/85 to track the pending `isClaim` issue.

PR #265, opened 2018-07-28
Refactor BatchReader

kennylavender:
I have not tested these changes yet to make sure they work, its almost 4 am. I am tired.

Here are some of the things this PR changes for the better

- Fixed the use of IOC
  Correct me if I am wrong but I believe this pr fixes an issue we had with IOC where the ClaimController directly imported the database implementation. This change uses IOC to remove the coupling between the ClaimController and the database. Database.ts is now a business logic interface that the DatabaseMongo.ts implements. Therefore the ClaimController no longer needs to import a Database implementation directly.

- Moves the mongo db connection logic out of the BatchReader.ts and into DatabaseMongo.ts

- Moves logic out of the Database Layer and back into the ClaimController

- Decomposed ClaimController.readNextBatch into a nice flow.
  - because the flow will now throw if an entry is not found, I was able to remove the entry check from Rotuer.ts

## Other possible improvements, but might save these for later.
- We could do the same with IPFS as we have done here with the Database, Storage.ts and StorageIPFS.ts etc.
- As we discussed before we could and should probably do something better with the Servcie logic, but I might save that later.

### PR Review Checklist


src/BatchReader/ClaimController.ts
```diff
@@ -1,30 +1,69 @@
...
+  private readonly findNextEntry: readFlow = async () => {
+    const readEntry = await this.db.readEntryFindIncomplete()
+    return { readEntry }
+  }
+
+  private readonly verifyEntryWasFound: readFlow = async x =>
+    x.readEntry ? Promise.resolve(x) : Promise.reject('No entries remaining')
+
+  private readonly incEntryAttempts: readFlow = async ({ readEntry, ...rest }) => {
+    const updatedEntry = { ...readEntry, attempts: readEntry.attempts + 1 }
```
kennylavender: Just realized this might be bad to increment attempts here due to race conditions. Handling this in mongodb would be better I think so we might need a specific function for this.

kennylavender:
Ended up taking this way further in this branch: https://github.com/poetapp/node/tree/refactor-batchreader-further

kennylavender:
Closing this idea PR. Once we decide how we want to refactor we will create new issues/prs.

PR #268, opened 2018-07-30
Use poet-js 2.0.0

lautarodragan:
This PR upgrades the version of poet-js to 2.0.0 and uses the `PoetTimestamp` interface provided by it.

It also adds some fixes required by poet-js 1.3.0.


lautarodragan:
Tests failing due to `isClaim` now returning false for values it previously returned true. Looking into it.

PR #282, opened 2018-08-01
Pagination suggestions

lautarodragan:


src/API/WorkController.ts
```diff
@@ -30,21 +29,19 @@ export class WorkController {
...
   async getById(id: string): Promise<any> {
     this.logger.trace({ method: 'getById', id }, 'Getting Work by Id from DB')
-    return this.collection.findOne({ id }, { fields: { _id: false } })
+    return this.collection.findOne({ id }, { projection: { _id: false } })
   }
 
-  async getByFilters(worksFilters: WorksFilters = {}): Promise<any> {
+  async getByFilters(worksFilters: WorksFilters = {}): Promise<ReadonlyArray<any>> {
     this.logger.trace({ method: 'getByFilters', worksFilters }, 'Getting Work by Filters from DB')
-    const { offset, limit }: { offset?: string; limit?: string } = omitUndefined(worksFilters)
```
lautarodragan: Here I didn't really find a reason to use `omitUndefined` to then spread-assign to offset and limit, since those would get default values afterwards, which would override `undefined`.

src/API/WorkController.ts
```diff
@@ -30,21 +29,19 @@ export class WorkController {
...
-    const skip = parseInt(offset, 10) || 0
-    const definedFilters = getDefinedFilters(worksFilters)
```
lautarodragan: I went ahead and removed this validation. `getByFilters` is assuming it's signature will be respected. (Unexpected fields in `WorksFilters` are not allowed and would thus break the build).

We can rest assured this will happen in run time since Joi is validating the parameters in the Router. If we wanted to be super careful, I think we could simply reuse the Joi schema rather than implementing a custom validator.

src/API/WorkController.ts
jwicks31: Ok that makes a lot of sense.

PR #298, opened 2018-08-06
Pagination counts

### LINKED ISSUE #276: Pagination Total Count
The node should send a header indicating the total count of elements available in `GET /works`.

This existed in the old node and is being used by some components in explorer-web which are not yet being used.

See for example https://github.com/poetapp/explorer-web/blob/11b8a8cd4b96baa6280ad5771c023925c771de89/src/components/organisms/WorksByProfile.tsx#L59-L64

```ts
  renderElement(works: ReadonlyArray<Api.WorksByFilters.Response>, headers: Headers) {
    const count = headers.get(Api.Headers.TotalCount) && parseInt(headers.get(Api.Headers.TotalCount), 10)

    if (!count) return this.renderNoWorks()
    else return this.renderWorks(works, count)
  }
```

We used to have a definition for this in poet-js: https://github.com/poetapp/poet-js/blob/4ccdcf3cf173dc4e9160a59e0726f732d6194162/src/Headers.ts

```ts
export namespace Headers {
  export const TotalCount = 'X-Total-Count';
  export const Unread = 'X-Unread';
}
```

We should add it again to poet-js and import it in the node.

TODO: create issue in poet-js

Needs https://github.com/poetapp/node/issues/95

jwicks31:
Resolves #276 

### Description of Changes
 - Adds "X-Total-Count" and "X-Unread" headers to getWorks
 - Exposes custom headers with KoaCors

I went back and forth between having separate functions inside of `getWorks` that returned the count, unread, and works fields individually. Since they same fields were reused in most of them I made getByFilters return all 3 in an object. I'm not really sure this is the best way, but I needed to use this in explorer-web, so I put it together.

### PR Review Checklist


src/API/WorkController.ts
```diff
@@ -13,6 +13,12 @@ interface WorksFilters {
   readonly limit?: number
 }
 
+interface WorkCount {
```
lautarodragan: `WorksWithCount`?

jwicks31:
@lautarodragan I made the changes and added an integration test as well.

test/Integration/GetWorks.ts
```diff
@@ -6,6 +6,11 @@ import { claimFromJSON } from 'Helpers/Claim'
 
 import { TheRaven, ABraveAndStartlingTrugh } from '../Claims'
 import { Client } from './Helper'
+export namespace Api {
```
lautarodragan: I think this is unused.

src/API/WorkController.ts
```diff
@@ -13,6 +13,11 @@ interface WorksFilters {
   readonly limit?: number
 }
 
+interface WorksWithCount {
+  readonly count: number
+  readonly works: ReadonlyArray<string>
```
lautarodragan: Why `ReadonlyArray<string>`? Aren't this something like `Work & { timestamp: PoetTimestamp }`? 

Might want to create a `WorkWithTimestamp` interface for that. It would be specific to the API module.

jwicks31: Honestly not really sure what I was thinking there.

geoffturk:
> The 'X-Total-Count' string and the WorksWithCount and WorkWithTimestamp interfaces will need to be moved to a Po.et Node Client in the future, most likely in poet-js, to de-duplicate code across projects, but we can address that in a separate PR, after the mainnet milestone.

@jwicks31 Can you create a ticket for this one so we have a record of it?

jwicks31:
@warrenv @kennylavender Whenever you get a chance I believe this one is ready to go.

src/API/WorkController.ts
```diff
@@ -32,17 +41,17 @@ export class WorkController {
...
+    const works = await this.collection
       .find(filters, { projection: { _id: false } })
-      .sort({
-        _id: -1,
-      })
+      .sort({ _id: -1 })
       .skip(offset)
       .limit(limit || 10)
       .toArray()
+    const count = await this.collection.count(filters)
```
warrenv: I'm wondering if this hits the database twice?
Maybe we could just get the results of find(), use length of that to get count
and then do works = results.sort().skip().limit().toArray()?

jwicks31: sort(), skip(), limit(), and toArray() are all mongodb methods, so I'm not sure that we could do it this way.

I tried and we can do it like this: 
```ts
const results = await this.collection
      .find(filters, { projection: { _id: false } })
    const resultsArray = await results.toArray()
    const count = resultsArray.length
    const works = await results
      .sort({ _id: -1 })
      .skip(offset)
      .limit(limit || 10)
      .toArray()
```
I don't think that would be any better and might be hitting the database 3 times?

jwicks31: `.count(filters)` was incorrect though. I updated it to `this.collection.find(filters, { projection: { _id: false } }).count()`

jwicks31:
@lautarodragan it dismissed your review whenever I updated the count function. Whenever you get a chance could you look again? The only change was updating count to `this.collection.find(filters, { projection: { _id: false } }).count()` from this.collection.count(filters)``

PR #302, opened 2018-08-07
Coverage, Unit Tests

wzalazar:
Fix https://github.com/poetapp/node/issues/301 https://github.com/poetapp/node/issues/308

### Description of Changes

This PR introduces:

- Adds unit test Extensions folder
- Adds `cache: false` in the `nyc` configuration
- Removes `lcov` reporter in the `nyc` configuration

### PR Review Checklist


src/Extensions/Array.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      given: 'array with two pairs of key values',
+      should: 'get the keys',
+      actual: Object.keys(result),
+      expected: ['key1', 'key2']
+    })
+
+    assert({
+      given: 'array with two pairs of key values',
+      should: 'get the values',
+      actual: Object["values"](result),
```
warrenv: Object.values(result) should work here. And it will be consistent with Object.keys() above.

wzalazar: yes, you're right, but Typescript doesn't like. I don't know why my IDE shows me this error.

`ts] Property 'values' does not exist on type 'ObjectConstructor'.`

But I will change it, If it pass all CI I think there will not be any problem

warrenv:
@lautarodragan Can you review again please? We'd like to get this merged today.

src/Extensions/Promise.test.ts
```diff
@@ -40,4 +41,18 @@ describe('Promise', async (should: any) => {
       expected: ERROR,
     })
   }
+
+  {
+    const actual = await Promise.reject(10)
+      .rethrow(error => error === 10 && TRANSLATION)
```
lautarodragan: Maybe add a ternary or `if` in the translate callback to make the intended behavior more explicit?

```
rethrow(error => error === 10 ? TRANSLATION : 'Something unexpected happened')
```

src/Extensions/Promise.test.ts
```diff
@@ -40,4 +41,18 @@ describe('Promise', async (should: any) => {
...
+      .then(success)
+      .catch(error => error)
+
+    assert({
+      given: 'Promise.reject(10).rethrow(10)',
```
lautarodragan: Maybe `given: 'Promise.reject(10).rethrow(TRANSLATION)', `?

src/Extensions/Promise.test.ts
```diff
@@ -40,4 +41,18 @@ describe('Promise', async (should: any) => {
...
+      should: 'reject',
```
lautarodragan: Maybe `should: 'reject with translated error`?

src/Extensions/Array.test.ts
```diff
@@ -0,0 +1,36 @@
+import { describe } from 'riteway'
+import './Array'
+
+describe('Array', async (should: any) => {
+  const { assert } = should('')
+
+  {
```
lautarodragan: Why are we unit testing JS' native functions?

src/Extensions/Array.test.ts
```diff
@@ -0,0 +1,36 @@
+import { describe } from 'riteway'
+import './Array'
+
+describe('Array', async (should: any) => {
```
lautarodragan: What do you think about explicitely grouping tests for this function?

`describe('Array.toObject', ...`

src/Extensions/Array.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    assert({
+      given: 'array with two pairs of key values',
+      should: 'transform to an object',
+      actual: typeof result === 'object',
+      expected: true
+    })
+
+    assert({
+      given: 'array with two pairs of key values',
+      should: 'get the keys',
```
lautarodragan: the object should have the same keys as the array?

src/Extensions/Array.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      actual: Object.keys(result),
+      expected: ['key1', 'key2']
+    })
+
+    assert({
+      given: 'array with two pairs of key values',
+      should: 'get the values',
```
lautarodragan: the object should have the same values as the array?

src/Extensions/Error.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe } from 'riteway'
+import './Error'
+
+describe('Extensions Error', async (should: any) => {
+  const { assert } = should('')
+
+  {
```
lautarodragan: Can you add tests for the behavior of `type` and `stack`?

src/Extensions/Array.test.ts
wzalazar: yeah, sounds good!

src/Extensions/Promise.test.ts
wzalazar: yep

src/Extensions/Promise.test.ts
wzalazar: yep

src/Extensions/Array.test.ts
wzalazar: do you disagree with the `should` text?

src/Extensions/Array.test.ts
```diff
@@ -0,0 +1,36 @@
+import { describe } from 'riteway'
+import './Array'
+
+describe('Array', async (should: any) => {
+  const { assert } = should('')
```
lautarodragan: I'm not sure whether this works as expected if we don't use the `should` from the immediate `describe`...

src/Extensions/Array.test.ts
lautarodragan: yeah, sorry. I was suggesting a text for the `should`.

src/Extensions/Error.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    })
+
+    assert({
+      given: 'a new Error',
+      should: 'get the keys of the new Error',
+      actual: Object.keys(actual),
+      expected: ['stack', 'message', 'type']
+    })
+
+    assert({
```
lautarodragan: There's a few more cases we may want to add if we want to be super thorough: the `type` property is set to the Error's constructor name if there isn't a set `type` property already.

So, for example, `new Error` will have `.type === 'Error'`, `new MyCustomError` will have `.type === 'MyCustomError` and if we do something like...

```ts
class MyCustomError extends Error {
  constructor(...) {
    this.type = 'SomeType'
  }
}
const err = new MyCustomError()
console.log(JSON.stringify(err)) // { ..., "type": "SomeType" }
```

This was done like this because MongoDB errors come with a `type` field while other errors don't.

And this only happens in the `toJSON` because native JS Error doesn't come with a decent `toJSON` method, and PinoJS (and other loggers) use the `.toJSON` method of passed objects to convert them to strings before sending them to the logging output.

Covering these cases will probably increase coverage, since the conditional in the toJSON may be considered as different branches of code.

It could also be nice to make explicit the fact that we're testing the `toJSON` method (in the cases we are). Unit tests also serve as good documentation (by declaring intent).

We can do this in a separate PR or open an issue if you prefer, either way works fine.

warrenv: Extending base objects in javascript is an anti-pattern and best avoided.
I would prefer to see us get the bare minimum of unit tests in place that will support a future refactoring away from these extensions.

lautarodragan: @warrenv I've been meaning to add documentation on why and how we're extending base JS objects. It's completely safe if done right.

We still may change a few of this in favor of composition with Ramda, but `Error.toJSON` in particular may live with us for a while. 

On the other hand, even if we stop binding these functions to base objects, we'll still have them, so we'll still want the unit tests.

src/Extensions/Array.test.ts
wzalazar: ok, I don't know, but just in case I will move it

src/Extensions/Error.test.ts
wzalazar: @lautarodragan added it!

wzalazar:
@warrenv can you review it?

PR #312, opened 2018-08-09
Unit test Bitcoin Helper

wzalazar:
Fix https://github.com/poetapp/node/issues/310 https://github.com/poetapp/node/issues/311

### Description of Changes

This PR introduces:

- unit test for Bitcoin
- unit test for asyncPipe **(deleted)**

### PR Review Checklist


kennylavender:
Just Fyi, I've been meaning to replace `asyncPipe` with ramda's `pipeP`. I am adding ramda in #317, then we can do that replacement.

src/Helpers/AsyncPipe.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe } from 'riteway'
+import { asyncPipe } from './asyncPipe'
+
+describe('asyncPipe', async (should: any) => {
+  const { assert } = should('')
+
+  {
+    const inc = (number = 0) => Promise.resolve(number + 1)
+    const actual = await asyncPipe(inc)(0)
```
kennylavender: I think it would be good to have a test with 3 or more functions in the pipe to make sure data flows correctly. Here are a couple of function ideas, double, and add10.

kennylavender:
In the future, it would be good to have these 2 changes in two separate PR's. PR's should have a single responsibility if possible. The reason behind that is, say that one of these needs changes made, then it starts holding back the other change from being merged in even though its good to go. And that change being held back could end up holding back more work. :D

src/Helpers/AsyncPipe.test.ts
lautarodragan: @kennylavender I'd go with merging your PR asap and using Ramda's pipeP as you commented instead, and avoid investing time and effort into testing asyncPipe. What do you think?

src/Helpers/Bitcoin.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  
+    return new bitcore.Transaction()
+      .from(utxo)
+      .to('1Gokm82v6DmtwKEB8AiVhm82hyFSsEvBDK', 15000)
+      .addData(new Buffer(`${prefix}0000ipfsDirectoryHash`))
+      .sign(privateKey);
+}
+
+describe('Bitcoin', async (should: any) => {
+  const { assert } = should('')
```
lautarodragan: @kennylavender why are we always passing empty string to should? I've seen this in every PR but I'm not sure whether this is the way to do it

src/Helpers/Bitcoin.test.ts
```diff
@@ -0,0 +1,81 @@
+import { describe } from 'riteway'
+import * as bitcore from 'bitcore-lib'
+import { getPoetTimestamp } from './Bitcoin'
+
+const createTransaction = (prefix: string) => {
+    // All data is a mock
+    const privateKey = new bitcore.PrivateKey('L1uyy5qTuGrVXrmrsvHWHgVzW9kKdrp27wBC7Vs6nZDTF2BRUVwy');
+    const utxo = [new bitcore.Transaction.UnspentOutput({
```
lautarodragan: @wzalazar why isn't tslint rejecting these double-quotes? are tests not linted?

src/Helpers/AsyncPipe.test.ts
kennylavender: @lautarodragan yup I agree :D

src/Helpers/Bitcoin.test.ts
kennylavender: @lautarodragan The should call provides a default *should* value for all the the assertions. Its not necessary to provide a string if there is not a common *should* between the assertions. I have a habit of typing an empty string. There are no issues with passing empty strings as far as I know, but if we notice empty strings being passed to should it would probabaly be good to remove them for cleanliness and to avoid any confusion.

lautarodragan:
@wzalazar can that be fixed?

wzalazar:
@lautarodragan yes, we can add this files. Just we have to create one task for this issue. However, I ran the task for this PR

src/Helpers/AsyncPipe.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe } from 'riteway'
+import { asyncPipe } from './AsyncPipe'
+
+describe('asyncPipe', async (should: any) => {
```
lautarodragan: Let's remove this unit test since we're adding rambda today

wzalazar: ok, I'm going to remove it now

kennylavender:
@wzalazar ramda was merged and I created a pr to replace asyncPipe, #322

PR #318, opened 2018-08-11
Generate SCREAMING_SNAKE_CASE from Config

### LINKED ISSUE #196: Support all configuration options via env vars
Currently some configuration options are only supported via configuration file.

https://github.com/poetapp/node/blob/f5e6ef07d93d7543bc430069e31bd066c8a341a2/src/Configuration.ts#L88-L102

We need to update `loadConfigurationFromEnv` to programmatically generate keys from values by applying camelCase to SCREAMING_SNAKE_CASE.

We could do this by iterating over the keys of `defaultConfiguration` if it contained every possible configuration key, which is not the case right now since some of them are optional.

jwicks31:
Resolves #196 

### Description of Changes

- Added Configuration Helper Function: `keysToScreamingSnakeCase` (open to name changes)
- Updated `loadConfigurationFromEnv` to use `keysToScreamingSnakeCase()`
- Added missing defaultConfiguration keys as undefined in defaultConfiguration
- Unit Tests

### PR Review Checklist


jwicks31:
@lautarodragan I added the configuration keys that were missing in defaultConfiguration as undefined values in defaultConfiguration. I was questioning if that was OK though.

src/Helpers/Configuration.ts
```diff
@@ -0,0 +1,6 @@
+import 'Extensions/Array'
+
+export const keysToScreamingSnakeCase = (object: object = {}): {} =>
+  Object.entries(object)
+    .map(([key]) => [key.replace(/([A-Z])/g, capital => '_' + capital).toUpperCase(), key])
+    .toObject()
```
kennylavender: After we get ramda package added to the node in #317, you can use [fromPairs](https://ramdajs.com/docs/#fromPairs) instead of using the custom Array method `.toObject`

src/Helpers/Configuration.ts
```diff
@@ -0,0 +1,6 @@
+import 'Extensions/Array'
+
+export const keysToScreamingSnakeCase = (object: object = {}): {} =>
+  Object.entries(object)
+    .map(([key]) => [key.replace(/([A-Z])/g, capital => '_' + capital).toUpperCase(), key])
```
kennylavender: It think it might be good to separate out a function that does this camel case to screaming snake case.

kennylavender: Then we could compose together the function something like this

```js
import { _keys, fromPairs, map, pipe } from 'ramda'

import { camelCaseToScreamingSnakeCase } from 'Helpers/camelCaseToScreamingSnakeCase'

const toPair = (s) => [camelCaseToScreamingSnakeCase(s), s]

export const keysToScreamingSnakeCase = pipe(_keys, map(toPair), fromPairs)
```

kennylavender: Also, I think there is probably a better name for this function, `keysToScreamingSnakeCase` sounds like we are converting the keys and the values remain the same previous values. What we are really doing is creating a map from ScreamingSnakeCase keys to CamelCase keys. There isn't really a great name that I can think of, but maybe something like `createKeyMapScreamingSnakeCaseToCamelCase`

jwicks31:
@kennylavender Ok I made the changes. Just waiting on #317 to be merged to update the rest.

jwicks31:
@kennylavender my mistake on that. I updated the entire function to use the ramda provided functions you mentioned above. I think it is better now and all unit tested.

src/Helpers/Configuration.test.ts
```diff
@@ -0,0 +1,51 @@
+import { describe } from 'riteway'
+import { createEnvToConfigurationKeyMap, toPair } from './Configuration'
```
kennylavender: We should not be importing these functions from Configuration, let's move them to a different file, Helpers probably.

src/Helpers/Configuration.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+describe('toPair()', async (should: any) => {
```
kennylavender: We don't need to test this `toPair` function, it's just an implementation detail of the main function. Lookup whitebox vs blockbox testing.

src/Helpers/Configuration.ts
```diff
@@ -0,0 +1,11 @@
+import { keys, fromPairs, map, pipe } from 'ramda'
+
+import { camelCaseToScreamingSnakeCase } from './camelCaseToScreamingSnakeCase'
+
+export const toPair = (s: string) => [camelCaseToScreamingSnakeCase(s), s]
```
kennylavender: don't need to export this to test, its just an implementation detail of `createEnvToConfigurationKeyMap`

src/Helpers/camelCaseToScreamingSnakeCase.test.ts
```diff
@@ -0,0 +1,27 @@
...
+describe('camelCaseToScreamingSnakeCase()', async (should: any) => {
+  const { assert } = should()
+
+  assert({
+    given: 'no arguments',
+    should: 'return an empty string',
+    actual: camelCaseToScreamingSnakeCase(),
+    expected: ''
+  })
+
```
kennylavender: Maybe add a `given empty string` test case, the other test cases look good :+1:

src/Helpers/Configuration.test.ts
jwicks31: It's importing this from `Helpers/Configuration`. Linting wouldn't let me do an absolute import. Should I rename the helper file?

src/Helpers/Configuration.test.ts
kennylavender: Eric mentions block box testing in this article a few times and explains it a bit https://medium.com/javascript-scene/mocking-is-a-code-smell-944a70c90a6a

kennylavender: The reason why I see it as a implementation detail is that its not general enough to be reused, in another function. Its very specific to the format of the function we are creating if that makes sense?

src/Configuration.ts
```diff
@@ -64,6 +66,11 @@ const defaultConfiguration: Configuration = {
   batchCreationIntervalInSeconds: 600,
 
   readDirectoryIntervalInSeconds: 30,
+
+  s3Url: undefined,
```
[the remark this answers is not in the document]
jwicks31: No. In the Issue @lautarodragan said: "We could do this by iterating over the keys of defaultConfiguration if it contained every possible configuration key, which is not the case right now since some of them are optional."

This was my attempt at fixing that. I didn't know what default values to assign to them though.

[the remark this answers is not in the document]
kennylavender: I see I think this works then if @lautarodragan is good with it. The one thing I might consider then is to make all properties required, not optional so that we don't accidentally forget or remove properties from the config.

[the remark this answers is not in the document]
lautarodragan: +1 to making them mandatory if it doesn't break anything

> No. In the Issue @lautarodragan said: "We could do this by iterating over the keys of defaultConfiguration if it contained every possible configuration key, which is not the case right now since some of them are optional."

This was a quick idea. This approach is fragile because some coder could unknowingly introduce a new configuration option to the interface but provide no default value for it. 

But I'm okay with that for now, as long as we are aware. Maybe we should leave a comment above the interface?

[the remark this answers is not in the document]
jwicks31: Made them mandatory and all integration tests are still passing. I also added the comment 
`
Provide default value in defaultConfiguration for any new configuration options
`
above the interface.

src/Helpers/camelCaseToScreamingSnakeCase.test.ts
jwicks31: I added this test.

src/Helpers/Configuration.ts
```diff
@@ -0,0 +1,11 @@
+import { keys, fromPairs, map, pipe } from 'ramda'
+
+import { camelCaseToScreamingSnakeCase } from './camelCaseToScreamingSnakeCase'
+
+const toPair = (s: string) => [camelCaseToScreamingSnakeCase(s), s]
+
+export const createEnvToConfigurationKeyMap = pipe(
```
lautarodragan: Since this function only cares about the values of the passed object, maybe it's best for it to actually receive the `keys` already, and let the caller have the responsibility (and freedom) of choosing _how_ to get those keys.

Also: since we don't have good typings for `pipe`, can you add a type signature to `createEnvToConfigurationKeyMap` indicating input and output type?

src/Configuration.ts
```diff
@@ -94,13 +102,7 @@ function loadConfigurationFromFile(configPath: string): Configuration | {} {
...
 
 function loadConfigurationFromEnv(): Partial<Configuration> {
-  // TODO: programmatically generate keys from values by applying camelCase to SCREAMING_SNAKE_CASE
-  const map: { [index: string]: string } = {
-    RABBITMQ_URL: 'rabbitmqUrl',
-    MONGODB_URL: 'mongodbUrl',
-    IPFS_URL: 'ipfsUrl',
-    INSIGHT_URL: 'insightUrl',
-  }
+  const map: { [index: string]: string } = createEnvToConfigurationKeyMap(defaultConfiguration)
```
lautarodragan: Best to properly type `createEnvToConfigurationKeyMap` and let TS infer the type of `map`.

src/Helpers/Configuration.ts
jwicks31: Ok cool. I've updated everything.

lautarodragan:
* [X] Tested changes manually
* [X] Checked accidental architectural/style changes
* [X] Reviewed entire diff
* [X] Unit tests
* [X] Documentation - not needed
* [X] Filenames and locations
* [ ] QE needed `@poetapp/quality-engineering Please review` 


Tested different combinations of `LOGGING_LEVEL=warn npm start`, `loggingLevel: "trace"` and default configuration. Seems to be working fine.

Not sure whether QE is needed. @poetapp/quality-engineering

PR #335, opened 2018-08-15
Unit Test Helper Time

wzalazar:
Fix https://github.com/poetapp/node/issues/334

### Description of Changes
 
This PR introduces 

- Unit test for the Helper Time

### PR Review Checklist


src/Helpers/Time.test.ts
```diff
@@ -0,0 +1,30 @@
+import { describe } from 'riteway'
+import { minutesToMiliseconds, secondsToMiliseconds } from './Time'
+
+describe('Time', async (should: any) => {
+  describe('minutesToMiliseconds', async (should: any) => {
```
kennylavender: Doing a nested describe format doesn't work because it's not truly nested. Check out the console output on success. `Time` is about in the middle of the screen with no tests. `minutesToMiliseconds` and `secondsToMiliseconds` are at the bottom.
screenshot from 2018-08-20 18-00-56

wzalazar: thanks @kennylavender !!! I will fix

wzalazar:
can you review again, @kennylavender?

PR #339, opened 2018-08-15
Unit Test Messaging

wzalazar:
Fix https://github.com/poetapp/node/issues/337

### Description of Changes

This PR introduces

- unit test for Messaging (is working progress)

### PR Review Checklist


src/Messaging/Messaging.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    assert({
+      given: 'start() with invalid connectionUrl',
+      should: `return the error: ${expected}`,
+      actual: await messaging.start().catch(e => e.message),
+      expected,
+    })
+  }
+
+  {
+    const { connect, assertExchange, publish } = createStubConnect()
```
lautarodragan: I'm sorry but I don't see how these unit tests add value. They look more complex than the code being tested — which is 99.9% amqplib and 0.01% our own code.

They also seem to focus too much on testing that the functions are called with this or that parameter (which is an implementation detail) rather than checking that they actually work.

Communication across modules should be tested with integration tests IMO. ([And it already is](https://github.com/poetapp/node/blob/master/test/Integration/PostWork.ts#L42))

PR #342, opened 2018-08-17
test: Unit Test HttpExceptionsMiddleware

wzalazar:

### Release

## Description of Changes

Fix https://github.com/poetapp/node/issues/341

This PR introduces 

- unit test for HttpExceptionsMiddleware
- refactor


src/API/Middlewares/HttpExceptionsMiddleware.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      const ctx = { status: 0, body: '' }
+      const next = () => Promise.reject(new Error('Something happened!'))
+
+      await HttpExceptionsMiddleware(ctx, next).catch()
+
+      assert({
+        given: 'an error',
+        should: 'return context with status 503 and the text Something happened!',
+        actual: ctx.status === 503 && ctx.body === 'Something happened!',
+        expected: true,
```
warrenv: The error report would be more readable if we actually compared objects.
```javascript
await HttpExceptionsMiddleware(ctx, next).catch()
const actual = { status: ctx.status, body: ctx.body }
const expected = { status: 503, body: 'Something happened!' }
```

src/API/Middlewares/HttpExceptionsMiddleware.ts
```diff
@@ -1,23 +1,33 @@
...
+
+const isIllegalArgumentException = (exception: Error) => exception instanceof IllegalArgumentException
+const isNotFoundException = (exception: Error) => exception instanceof NotFoundException
+
+const getIllegalArgumentException = ({ message }: Error) => ({ status: 422, body: message })
+const getNotFoundException = ({ message }: Error) => ({ status: 404, body: message })
+const getDefaultException = ({ message }: Error) => ({ status: 503, body: message })
+
+export const getCodeError = (exception: Error): { body: string; status: number } =>
+  cond([
```
warrenv: nice use of ramda!

src/API/Middlewares/HttpExceptionsMiddleware.test.ts
```diff
@@ -0,0 +1,99 @@
+import { IllegalArgumentException, NotFoundException } from '@po.et/poet-js'
+import { describe } from 'riteway'
+import { spy } from 'sinon'
+import { getCodeError, HttpExceptionsMiddleware, executeLogger } from './HttpExceptionsMiddleware'
+
+describe('HttpExceptionsMiddleware', async (should: any) => {
+  describe('HttpExceptionsMiddleware middleware', async (should: any) => {
```
kennylavender: remove nested describes, the format does not work with riteway. Check the test output.

wzalazar:
can you review again, @kennylavender?

src/API/Middlewares/HttpExceptionsMiddleware.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  const { assert } = should('')
+  {
+    const ctx = {}
+    const next = spy()
+    await HttpExceptionsMiddleware(ctx, next)
+
+    assert({
+      given: 'no errors',
+      should: 'call next once',
+      actual: next.calledOnce,
```
lautarodragan: Dunno how this works, but doesn't `next` get called every time no matter what?

```ts
  try {
    await next()
  } catch ...
```

Maybe we should test that `context.body` and `context.status` were not mutated by `HttpExceptionsMiddleware`?

wzalazar: well, just calling the spy you are securing the `next` resolve with success. But I can add the check the immutability the `ctx` also

wzalazar: done!

lautarodragan: Thanks, but my point is that `next.calledOnce === true` doesn't mean there are _no errors_, which is what we're saying in the `given` clause.

`.next` gets always *always*, errors or no errors, so this unit test doesn't really test anything.

I wrote an small example of this: https://github.com/poetapp/node/pull/375

src/API/Middlewares/HttpExceptionsMiddleware.test.ts
```diff
@@ -0,0 +1,114 @@
...
+})
+
+describe('HttpExceptionsMiddleware executeLogger()', async (should: any) => {
+  const { assert } = should('')
+  {
+    const context = { logger: { error: spy() } }
+    const exception = new Error()
+    const status = 200
+
+    executeLogger(context, exception, status)
```
lautarodragan: We can avoid testing this side effect (calling the logger) and increase readability a bit by introducing another mini refactor. See related comment.

src/API/Middlewares/HttpExceptionsMiddleware.ts
```diff
@@ -1,23 +1,33 @@
...
+    context.status = status
 
-      context.body = 'Internal Server Error'
-      context.status = 503
-    }
+    executeLogger(context, exception, status)
   }
 }
+
+export const executeLogger = (context: any, exception: Error, status: number) => {
```
lautarodragan: Isn't the `status` parameter a duplication of `context.status` here?

Maybe we can avoid it completely.

src/API/Middlewares/HttpExceptionsMiddleware.ts
```diff
@@ -1,23 +1,33 @@
...
+  if (context.logger && (status !== 422 && status !== 404))
```
lautarodragan: If we extract the `(status !== 422 && status !== 404)` to a function, we could unit test it, and it'd also give us readability by declaring intention in the function's name.

For example:

```ts
const isUnexpectedErrorStatusCode = (statusCode: number) => status !== 422 && status !== 404

export const executeLogger = (context: any, exception: Error) => {
  if (context.logger && isUnexpectedErrorStatusCode(context.status))
    context.logger.error({ exception }, 'Error Caught at Middleware - Will Return Internal Server Error')
}
```

src/API/Middlewares/HttpExceptionsMiddleware.ts
```diff
@@ -1,23 +1,33 @@
-import * as Koa from 'koa'
-
 import { IllegalArgumentException, NotFoundException } from '@po.et/poet-js'
+import { cond, T } from 'ramda'
 
-export const HttpExceptionsMiddleware: Koa.Middleware = async (context: any, next: () => Promise<any>) => {
```
lautarodragan: Instead of removing the type signature, try declaring the mocks of the context as `any` in the unit tests.

```ts
const ctx: any = {}
await HttpExceptionsMiddleware(ctx, next)
```

That way we'll keep the type-safety in the code that reaches production while avoiding the need to mock the entire context object for the integration tests.

wzalazar: I will try it :)

src/API/Middlewares/HttpExceptionsMiddleware.test.ts
wzalazar: I need to test the method `executeLogger ` if not I get this result 

```
src/API/Middlewares               |     92.5 |    64.29 |    90.91 |    93.94 |                   |
  HttpExceptionsMiddleware.ts     |    92.59 |    33.33 |    88.89 |    95.24 |                20 |
```

instead of:

```
src/API/Middlewares               |     97.5 |    92.86 |      100 |    96.97 |                   |
  HttpExceptionsMiddleware.ts     |      100 |      100 |      100 |      100 |                   |
```

src/API/Middlewares/HttpExceptionsMiddleware.ts
wzalazar: done

src/API/Middlewares/HttpExceptionsMiddleware.ts
wzalazar: fixed

src/API/Middlewares/HttpExceptionsMiddleware.ts
wzalazar: done

lautarodragan: I think there's an accidental extra white space there

wzalazar: yes, fixed ;)

src/API/Middlewares/HttpExceptionsMiddleware.test.ts
lautarodragan: @wzalazar what about this one?

wzalazar: I changed the given to `given: 'no errors or errors'`, but I'm not sure with it, what do you say?

lautarodragan: Even more confusing 😛 

If we remove the test, does the coverage drop?

wzalazar: yes, maybe is more confusing... ahahahah

yes, I can remove this test without suffering drop coverage

lautarodragan: Awesome 🚀

PR #347, opened 2018-08-20
Use await for creating claims

krobi64:

## Description of Changes

Fixes poetapp/frost-api#243 
Needs poetapp/poet-js#96

Modify node to treat `createClaim` as async


krobi64:
@lautarodragan This will fail until we release poet-js right?

lautarodragan:
Yep. Using `await` is a harmless change since anything can be awaited, but the change in the interface is breaking.

kennylavender:
@krobi64 can we get rid of the parse claim function inside here with this pr? It was added specifically because of the issue with the created/dateCreated being a Date object

https://github.com/poetapp/node/blob/master/src/Storage/ClaimController.ts#L205-L235

lautarodragan:
re package-lock conflict: a simple way to solve it is `git checkout master -- package-lock.json` and then `rm -r node_modules` and `npm i`.

lautarodragan:
Tagging `don't merge` for now since I think we're waiting for a reply to the review. @krobi64 lmk otherwise.

.travis.yml
```diff
@@ -10,6 +10,12 @@ services:
   - mongodb
   - rabbitmq
   - docker
+before_install:
+  - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
```
wzalazar: I have a doubt here if we are adding this package, will it be necessary to add the same in the docker image?
@WesleyCharlesBlake ping!

lautarodragan: good call @wzalazar — wondering the same now!

lautarodragan:
Alright, I tried upping the fee and now I'm getting `64: too-long-mempool-chain. Code:-26`.

I left it retrying for a long while and it never succeeded in broadcasting a single transaction.

I'm going to try and get the BlockchainWriter migrated to Bitcoin Core so we can test this PR in regtest.

.travis.yml
WesleyCharlesBlake: @wzalazar @lautarodragan [https://github.com/poetapp/node/blob/master/Dockerfile#L3] the only packages being installed from apt in the Docker image is `rsync`.

Do we need to include the PPA `ppa:ubuntu-toolchain-r/test` ? what are we installing from this PPA? is it only `gcc-5 g++-5` ?

If this is a dep, we can build it into the Docker image by adding directives to the Dockerfile

wzalazar: @WesleyCharlesBlake 
> Do we need to include the PPA ppa:ubuntu-toolchain-r/test ? what are we installing from this PPA? is it only gcc-5 g++-5 ?

yes, is a dependency for jsondl

for more context, I left this issue
https://github.com/digitalbazaar/jsonld.js/issues/254

Issue #234, opened 2018-07-17
Router in Storage swallowing errors


lautarodragan:
https://github.com/poetapp/node/blob/6a00eaff4ae8145fba9efff3d0ab8e5a271c8f13/src/View/Router.ts#L55-L61

All the functions being passed to `messaging.consume` may be throwing errors that may or may not get properly caught and handled.

They should be wrapper with `try` and `catch` and errors logged, as done in other routers:

https://github.com/poetapp/node/blob/6a00eaff4ae8145fba9efff3d0ab8e5a271c8f13/src/Storage/Router.ts#L53-L63

geoffturk:
@warrenv Reassigning to you. Feel free to farm it out. :-)

PR #349, opened 2018-08-21
logger errors

wzalazar:
Fix https://github.com/poetapp/node/issues/234

### Description of Changes

This PR introduces:

-  adds loggers in the view router

### PR Review Checklist

### DIFF
--- a/src/View/Router.ts
+++ b/src/View/Router.ts
@@ -39,20 +39,35 @@ export class Router {
   }
 
   onNewClaim = async (message: any) => {
-    const messageContent = message.content.toString()
+    const logger = this.logger.child({ method: 'onNewClaim' })
 
-    await this.workController.createWork(JSON.parse(messageContent))
+    const messageContent = message.content.toString()
+    logger.info({ messageContent }, 'Setting message content on works')
+
+    try {
+      await this.workController.createWork(JSON.parse(messageContent))
+    } catch (error) {
+      logger.error({ error }, 'Failed to create on works')
+    }
   }
 
   onClaimIPFSHash = async (message: any) => {
+    const logger = this.logger.child({ method: 'onClaimIPFSHash' })
+
     const messageContent = message.content.toString()
     const { claimId, ipfsFileHash } = JSON.parse(messageContent)
 
-    await this.workController.setIPFSHash(claimId, ipfsFileHash)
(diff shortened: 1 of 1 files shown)


src/View/Router.ts
```diff
@@ -39,20 +39,35 @@ export class Router {
   }
 
   onNewClaim = async (message: any) => {
+    const logger = this.logger.child({ method: 'onNewClaim' })
+
     const messageContent = message.content.toString()
+    logger.info({ messageContent }, 'Setting message content on works')
```
lautarodragan: Why did you add this message at `info` level?

src/View/Router.ts
```diff
@@ -39,20 +39,35 @@ export class Router {
...
+    try {
+      await this.workController.setIPFSHash(claimId, ipfsFileHash)
+    } catch (error) {
+      logger.error({ error }, 'Failed to set claimId and ipfsFileHash on works')
+    }
   }
 
   onBatchWriterCreateNextBatchSuccess = async (message: any): Promise<void> => {
-    const logger = this.logger.child({ method: 'onBlockchainWriterRequestTimestampRequest' })
+    const logger = this.logger.child({ method: 'onBatchWriterCreateNextBatchSuccess' })
```
lautarodragan: Nice catch

src/View/Router.ts
```diff
@@ -81,22 +96,32 @@ export class Router {
   }
 
   onIPFSHashTxId = async (message: any) => {
+    const logger = this.logger.child({ method: 'onIPFSHashTxId' })
+
+    logger.info({ message }, 'Message')
```
lautarodragan: Same question here - why is this message being added, and at info level?

src/View/Router.ts
wzalazar: well, just I follow the same pattern that I saw here

https://github.com/poetapp/node/blob/b2d4180edb9b0018f9faf0a5cd3f126c181c7da4/src/View/Router.ts#L127-L129

I thought you were logging the inputs before of these been used in the call

what do you say?

lautarodragan: Got it.

I haven't formally defined what logging levels we should use in each scenario yet, so for now can you change them to `trace` to be on the safe side?

I don't want the `info` level to be too verbose.

Issue #351, opened 2018-08-22
Unit test Exceptions


wzalazar:

wzalazar:
closed with https://github.com/poetapp/node/pull/352

PR #352, opened 2018-08-22
Storage Exceptions, unit tests

wzalazar:

## Description of Changes

Fix https://github.com/poetapp/node/issues/351

This PR introduces:

- unit test for Storage Exceptions

### DIFF
--- a/src/Storage/Exceptions.test.ts
+++ b/src/Storage/Exceptions.test.ts
@@ -1,0 +1,148 @@
+import { pick } from 'ramda'
+import { describe } from 'riteway'
+import '../Extensions/Error'
+import { FetchErrorType } from '../Helpers/FetchError'
+import { FailureReason } from './DownloadFailure'
+import {
+  NoMoreEntriesException,
+  InvalidClaim,
+  IPFSGenericError,
+  IPFSTimeoutError,
+  errorToIPFSError,
+} from './Exceptions'
+
+class FetchError extends Error {
+  readonly type: string
+
+  constructor() {
+    super()
+    this.name = 'FetchError'
+    this.type = FetchErrorType.BodyTimeout
+  }
+}
+
+describe('Exceptions NoMoreEntriesException', async (should: any) => {
+  const { assert } = should('')
+
+  {
+    const noMoreEntriesException = new NoMoreEntriesException('NoMoreEntriesException')
+
+    assert({
+      given: 'the new instance of NoMoreEntriesException',
+      should: 'be an instance of NoMoreEntriesException',
+      actual: noMoreEntriesException instanceof NoMoreEntriesException,
+      expected: true,
+    })
+
+    const actual = JSON.parse(JSON.stringify(noMoreEntriesException)).message
+    const expected = 'NoMoreEntriesException'
+
+    assert({
+      given: 'the new instance of NoMoreEntriesException',
+      should: 'return the message NoMoreEntriesException',
+      actual,
+      expected,
+    })
+  }
+})
+
+describe('Exceptions InvalidClaim', async (should: any) => {
+  const { assert } = should('')
+
+  {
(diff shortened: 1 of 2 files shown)


src/Storage/Exceptions.test.ts
```diff
@@ -0,0 +1,148 @@
...
+describe('Exceptions InvalidClaim', async (should: any) => {
+  const { assert } = should('')
+
+  {
+    const invalidClaim = new InvalidClaim('ipfsFileHash', FailureReason.InvalidClaim)
+
+    assert({
+      given: 'the new instance of InvalidClaim',
+      should: 'be an instance of InvalidClaim',
+      actual: invalidClaim instanceof InvalidClaim,
```
warrenv: Doesn't this just test the 'new' keyword? I don't think these types of tests are needed.

wzalazar: but it is a use case, for example here is used 

https://github.com/poetapp/node/blob/00f92a6afd7f605ee1c2a10cfe8727970e09216c/src/Storage/ClaimController.ts#L120-L129

warrenv: ok, but keep in mind that using 'instanceof' has problems:
https://medium.com/javascript-scene/common-misconceptions-about-inheritance-in-javascript-d5d9bab29b0a

wzalazar: thanks, @warrenv for remembering it, well maybe is a potential issue, a good candidate to refactor

lautarodragan: For reference, `instanceof` problems have been fixed in new versions of JavaScript. See [hasInstance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance).

src/Storage/Exceptions.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+}
+
+describe('Exceptions NoMoreEntriesException', async (should: any) => {
+  const { assert } = should('')
+
+  {
+    const noMoreEntriesException = new NoMoreEntriesException('NoMoreEntriesException')
+
+    assert({
```
lautarodragan: This test is redundant — you have just declared `noMoreEntriesException` to be an instance of `NoMoreEntriesException` in the test.

src/Storage/Exceptions.test.ts
lautarodragan: I agree with Warren here. Same as above, even though the code is using `instanceof`, this test isn't really testing that. It's just testing the `new` keyword.

To actually test that logic you'd need to refactor the code, extracting the logic into a pure function that takes an `error` and returns an object `{ipfsFileHash: string, failureType: FailureType, failureReason: 
 FailureReason }`, returning `undefined` instead of `else throw error` and handling the side-effect `logger.warn({ error }` from the calling function.

If you like that approach, let's do it in a separate PR so we can sooner merge this one.

src/Storage/Exceptions.test.ts
```diff
@@ -0,0 +1,148 @@
...
+})
+
+describe('Exceptions IPFSGenericError', async (should: any) => {
+  const { assert } = should('')
+
+  {
+    const underlyingError = new Error('underlyingError')
+    const ipfsGenericError = new IPFSGenericError('ipfsFileHash', underlyingError)
+
+    assert({
```
lautarodragan: Same as above, just testing the `new` keyword.

src/Storage/Exceptions.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+})
+
+describe('Exceptions IPFSTimeoutError', async (should: any) => {
+  const { assert } = should('')
+
+  {
+    const ipfsTimeoutError = new IPFSTimeoutError('ipfsFileHash')
+
+    assert({
```
lautarodragan: Same as above, just testing the `new` keyword.

PR #353, opened 2018-08-23
Update package.json and LICENSE

geoffturk:

## Description of Changes

Update `package.json` and `LICENSE` files with MIT license and other changes.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@po.et/node",
   "version": "1.0.3",
-  "description": "",
+  "description": "The Po.et Node allows you to timestamp documents in a decentralized manner. It is built on top of the Bitcoin blockchain and IPFS.",
```
warrenv: Should we use 'anchor'?

geoffturk: Good catch!

geoffturk: Thinking about it some more, I think "timestamp" is OK here. Using "anchor" seems a little strange in this context:
"The Po.et Node allows you to anchor documents in a decentralized manner. It is built on top of the Bitcoin blockchain and IPFS."

I would only use it if we changed it to:
"The Po.et Node allows you to anchor documents to the Bitcoin blockchain."

But "timestamp documents in a decentralized manner" sounds OK as is.

PR #230, opened 2018-07-16
Coverage report

wzalazar:
### Description of Changes

Introduces report of coverage for unit test and integration test.

I run the coverage separated because are different instances and when you run another after the first one it will rewrite the input. Maybe when we have the all test in the same instance we will be able to capture the all the coverage together.

needed-by https://github.com/poetapp/node/issues/11

### PR Review Checklist

* [ ] **Have you tested the changes?**
* [ ] **Is QE review required? If so, has this been reviewed by QE?** (For all new features, review by QE is required. Minor changes which already have passing functional tests may not need a new QE review -- when in doubt, loop in a QE reviewer).
* [ ] **Are there code changes that require new unit tests? If so, are those tests included?** (A separate PR for tests is not appropriate).
* [ ] **Are documentation updates required? If so, does this PR include it?** (Separate documentation PR is generally not appropriate).
* [ ] **Are the added/edited files named appropriately and located correctly in the project directory structure?**

PR #364, opened 2018-08-24
Extract StorageWriter from Storage

### LINKED ISSUE #363: Extract StorageWriter from Storage
part of #360 

Extracting just the StorageWriter, without renaming Storage to StorageReader allows a much smaller PR that should be easier to review.

kennylavender:

## Description of Changes

closes #363

Extracts StorageWriter from Storage without renaming Storage to StorageReader. Renaming StorageReader in this PR would have created a ton of changed files.

Does not modify the code from StorageWriter or Storage in any away except for extraction.


kennylavender:
Update: insight api still down it seems.

kennylavender:
I've spotted a bug in the Storage module, this bug is in master as well. Created an issue here #369

I've tested this code change and it is working beside the bug described above and that will be fixed in another issue.

Ready for review @lautarodragan @poetapp/quality-engineering

src/StorageWriter/ClaimController.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      new Db('poet', server),
+      new Messaging(),
+      new IPFS(IPFSConfiguration),
+    )
+
+    assert({
+      given: 'the new instance of ClaimController',
+      should: 'be an instance of ClaimController',
+      actual: claimController instanceof ClaimController,
+      expected: true,
```
warrenv: This just seems to be testing the 'new' keyword. What situation could ever cause instanceof to be false?

kennylavender: This is code introduced by this PR https://github.com/poetapp/node/pull/230. Not in the scope of this PR. I agree this could be improved, IMO this test is really testing that it does not throw an error, so the `should` could be updated. But this should be discussed in a separate issue.

warrenv: Thanks. Have a look at using 'git mv' when renaming files if you are not familiar with it - it will prevent the renamed file from showing up as new code. I will say it's hard to get into the habit especially since renaming files is often done in IDEs, but makes PR review easier.

kennylavender: Thanks, I am splitting a single module into 2 modules here so `git mv` would not work for this case.

lautarodragan:
Insight still down...

Code looks good. Haven't checked tests yet.

kennylavender:
@lautarodragan yeah I was getting nothing but cloud flare errors from Insight the first night I tried, the second night I tried I was still getting them, but some of the fetches made it through, I just let it run until it finally pulled the block needed.

PR #377, opened 2018-08-28
Fix downloadSuccessTime in Storage module

### LINKED ISSUE #369: Storage Module - downloadSuccessTime not being set on entry
Spotted a bug in the Storage Module

The method downloadEntryClaim in the ClaimController is no longer setting downloadSuccessTime value on the entries. ~which will result in entries being continually redownloaded until max attempts have been reached.~

kennylavender:

## Description of Changes

closes #369 

- Fix the ClaimController to set the `downloadSuccessTime` value on the entries after successful downloads.
- Fix the `downloadSuccessTime` property on the entry interface


kennylavender:
@lautarodragan @poetapp/quality-engineering ready for review, I finally got to test this manually while insight api was working for a few minutes.

kennylavender:
@warrenv I have now removed the only unit testable code change as it was unnecessary, I almost did not add it to begin with.

src/Storage/ClaimController.ts
```diff
@@ -234,6 +235,32 @@ export class ClaimController {
...
+    )
+
+    logger.trace('finished setting download success time')
+
+    return {
+      entry,
+      ...rest,
+    }
+  }
+
```
warrenv: This function is doing two things: adding a downloadSuccessTime prop if it doesn't exist in it's argument AND updating the downloadSuccessTime in the database (whether it was missing in 'entry' or not).

That first part is a piece of conditional logic that can be extracted and unit tested. 
Something like this (not completely sure about the mongo call):

```javascript
  const updateWithCurrentTime = obj => Promise.resolve({ downloadSuccessTime: new Date().getTime(), ...obj})

  const updateDatabase = async (entry) => {
    await this.collection.updateOne(
      {
        _id: entry._id,
      },
      { $set: { downloadSuccessTime: entry.downloadSuccessTime } }
    )

    return entry
  }

  public setDownloadSuccessTime = pipeM(updateWithCurrentTime, updateDatabase)
```

The `updateWithCurrentTime` function would have 2 unit tests to handle the conditional:
  - given an object without a dowloadSuccessTime prop
    should return an object with the prop added
  - given an object with a downloadSuccessTiome prop
    should return an the same object.

`updateWithCurrentTime` could even be generalized further by taking the prop name if it would be useful in other areas of the codebase.

kennylavender:
@warrenv made the change we discussed.

src/Storage/Entry.ts
```diff
@@ -3,6 +3,6 @@ export interface Entry {
   readonly ipfsFileHash: string
   readonly claimId?: string
   readonly lastDownloadAttempt?: number
-  readonly lastDownloadSuccess?: number
+  readonly downloadSuccessTime?: number
```
wzalazar: I have a doubt here, I found `downloadSuccessTime` in several places with a `null` as a default value. Is it ok?

in this file https://github.com/poetapp/node/blob/de7c15ec963a8e07796a654e0c3627470edce167/src/Storage/ClaimController.ts

src/Storage/ClaimController.ts
wzalazar: I agree with @warrenv if we can use function composing cool

kennylavender: @warrenv and I already discussed this via zoom. What he proposed only works if I am modifying the pass-through data which I am not anymore. I removed that logic from this PR.

src/Storage/Entry.ts
kennylavender: Good catch. I did not think about that.

I think because the property is optional it is ok, we are using null to make sure the mongo database has the property, where if we passed undefined to mongo I am pretty sure mongo would ignore the property.

Given that this was pre-existing code that was accidentally removed at some point. I think this is something we should come back to after mainnet and maybe we can come up with better ways of working with mongo flows and data.

kennylavender:
@wzalazar @warrenv I have been wanting to look into using ramda `over` to see if we can compose the flows of the node better, but the problem is I don't believe `over` will work with Promises. I think its possible to compose better using `over` but I think it might require a `Task` monad or something. I just don't have time to research this stuff before mainnet.

kennylavender:
@lautarodragan I believe your right, I've updated the issue, I think that might have me jumping to a conclusion before testing if the missing `lastDownloadAttemptTime` was causing redownloads.

Issue #106, opened 2018-05-17
Use bitcoind - BlockchainReader


geoffturk:
BlockchainReader can just getblock :blockId.

Split out from #24.

geoffturk:
Needs #156

Issue #24, opened 2017-12-05
Use bitcoind + bitcoinjs-lib


lautarodragan:
Drop Insight API and Bitcore altogether.

BlockchainWriter will require bitcoinjs-lib for transaction building.

Module will need to use bitcoind's json-rpc, and since Bitcoin Core doesn't index utxos, we'll need it to be wallet-enabled too.

Add bitcoind docker image for development and local testing, running in regtest by default.

### Prior Research

- https://github.com/poetapp/bitcoind-docker
- https://github.com/poetapp/bitcoind-client

lautarodragan:
https://github.com/poetapp/bitcoind-client

lautarodragan:
Please wait a bit before starting work on this.

geoffturk:
BlockchainReader split out into #106.

geoffturk:
Research alternatives to bitcoinjs-lib first to see if there is a better option.

wzalazar:
we have these repositories related, it's a before research

https://github.com/poetapp/bitcoind-docker
https://github.com/poetapp/bitcoind-client

krobi64:
What doin you mean by Bitcoin Core does not index utxos? From https://bitcoin.stackexchange.com/questions/37397/where-is-the-utxo-data-stored

> For the tip (the last block in the active chain), we maintain a database with the set of unspent transaction outputs (UTXOs), stored in $DATADIR/chainstate, and aggressively cached in memory. When a new block is received, we remove the UTXOs spent by transactions in the block, and add the outputs of those transactions. When a reorganization happens, we use undo data (stored in $DATADIR/blocks/rev*.dat) to roll back the changes to the UTXO set. In a way, you can see blocks as patches to the UTXO set - but the UTXO set is never explicitly committed to or stored in blocks, it is merely implied by them.

lautarodragan:
@krobi64 I may be wrong in that statement. I also believe whether the entire UTXO set is accessible or just the ones for wallets is configurable.

But I'm not 100% sure of any of this. What I did find out is that when I tried to use the get UTXO RPC call on any address I'd get an empty array in return, unless the account had been created and managed in that bitcoind. 

I then did some reading on the subject and found people mentioning how not the entire UTXO set was stored, and that being one of the reasons things like bitcore existed. But all of that could be outdated info.

If you can get the UTXO of any address that would allow us to continue creating transactions are we currently are — otherwise we'd need to use the wallet RPC calls.

geoffturk:
[...]

Issue #22, opened 2017-12-05
Travis Node Sync Testing


lautarodragan:
We can configure Travis to run 2 nodes inside docker containers. 

### We'll need
- COPY dist/babel, package.json, etc., to ensure we're running the code being pushed and that the containers are isolated from each other.
- Run two bitcoind containers in -regtest, and `addnode` each other.
- Run two IPFS containers.
- Run two MongoDBs. This one is easy: just use two different connection URLs: `mongodb://localhost:27017/poet1` and `mongodb://localhost:27017/poet2`.
- Run two RabbitMQs. We're currently using Travis' `services` for this - we'll have to check whether Travis supports running more than one of the same service or use a docker container otherwise. 

Then we'll have to build some special integration tests to test replication across nodes. We could have some environment variables that are picked up by integration tests, such as `SECONDARY_NODE_URL`.

patrickleet:
you may want to consider docker-compose and jenkins instead of travis

warrenv:
Completed with https://github.com/poetapp/node/pull/545

PR #383, opened 2018-08-30
Replace Insight with Bitcoin Core in BlockchainReader

lautarodragan:

Fixes https://github.com/poetapp/node/issues/106
Part of https://github.com/poetapp/node/issues/24

RDD: [Replacing Insight API with Bitcoin Core](https://github.com/poetapp/rdd/blob/master/replace-insight-with-bitcoind.md).

## Description

This PR completely drops the Insight API dependency for the BlockchainReader module, replacing it with Bitcoin Core.

The Helpers/Bitcoin file now only contains interfaces and enums used with Bitcoin Core's RPC, and the declaration of the two prefixes we support.

Bitcoin-related functions that are specific to the BlockchainReader now live in BlockchainReader/Bitcoin.

## New Configuration Options

Configuration options for Bitcoin Core RPC have been added.

## New Dependencies

I'm using Rui Marinho's [bitcoin-core](https://github.com/ruimarinho/bitcoin-core), a simple wrapper over the Bitcoin Core RPC, with custom TypeScript types for it.

186 repositories currently [depend on ruimarinho/bitcoin-core](https://github.com/ruimarinho/bitcoin-core/network/dependents).

### Old Tests

Two test files needed to be updated to reflect the changes, but I decided to delete them instead. They weren't really testing anything — they were originally added to circumvent the coverage report failing to correctly calculate coverage for TypeScript files.

### New Unit Tests

Almost all logic is written in pure functions in BlockchainReader/Bitcoin and unit tested.

I've added a real bitcoin block as json for this purpose.

### Functional Tests

This change can only be correctly tested with [Full Node Syncing](https://github.com/poetapp/node/issues/22), which is blocked by this PR.

### Docker Image of Bitcoin Core

I've added a bitcoin-core entry to docker-compose.yml.

I'm running this with `sudo docker-compose up -d poet-bitcoind`.

It's configured to match the node's default configuration.

### Manual Testing

- Set `"forceBlockHeight": 1411304`
- `db.dropDatabase()`
- `npm start`
- `GET http://localhost:18080/works` should return an empty array
- Let it run for a while
- `GET http://localhost:18080/works` should now return 6 works, each with a full timestamp.
- Compare results of `GET http://localhost:18080/works` to [block-1411304.json.tar.gz](https://github.com/poetapp/node/files/2335054/block-1411304.json.tar.gz)

More and different manual testing is welcome.

## Future Changes

We're currently passing `username` and `password` to the Bitcoin Core RPC. This method is unsafe and deprecated, soon to be removed. We should use the `auth` option instead, which supports passing the hash of the password rather than plaintext.

We'll address this change in a separate PR.


src/BlockchainReader/ClaimController.ts
```diff
@@ -15,55 +17,48 @@ export class ClaimController {
...
-        blockHeight,
-        blockHash,
-      }))
-    const transactionIds = block.transactions.map(_ => _.id)
+    const block: Block = await this.bitcoinCore.getBlock(blockHash, GetBlockVerbosity.Transactions)
+
+    const poetAnchors = blockToPoetAnchors(block)
 
-    const matchingPoetTimestamps = poetTimestamps
+    const matchingPoetTimestamps = poetAnchors
```
lautarodragan: This part can be extracted into pure functions and unit tested. I'll work on it.

src/BlockchainReader/Bitcoin.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    const anchorIsVersion0003 = (poetAnchor: PoetTimestamp) => equals(poetAnchor.version, [0, 0, 0, 3])
+    const anchorIsBlockHash = (poetAnchor: PoetTimestamp) => poetAnchor.blockHash === TestBlock.hash
+    const anchorIsBlockHeight = (poetAnchor: PoetTimestamp) => poetAnchor.blockHeight === TestBlock.height
+    const anchorHasUnexpectedIpfsHash = (poetAnchor: PoetTimestamp) =>
+      !ipfsDirectoryHashes.includes(poetAnchor.ipfsDirectoryHash)
+
+    const anchorsIpfsHashes = poetAnchors.map(poetAnchor => poetAnchor.ipfsDirectoryHash)
+
+    assert({
+      given: 'blockToPoetAnchors(TestBlock)',
```
warrenv: These should read like descriptions. Something like:
- given:' a test block with poet anchors'

src/BlockchainReader/Bitcoin.ts
```diff
@@ -0,0 +1,53 @@
...
+
+interface VOutWithTxId extends VOut {
+  readonly transactionId: string
+}
+
+export const blockToPoetAnchors = (block: Block): ReadonlyArray<PoetTimestamp> =>
+  block.tx
+    .map(transactionToPoetAnchor)
+    .filter(_ => _)
+    .filter(poetAnchorHasCorrectPrefix)
```
warrenv: If this function could handle null entries the filter() above could be removed.
Perhaps transactionToPoetAnchor() should strip out any null records?

src/BlockchainReader/Bitcoin.ts
```diff
@@ -0,0 +1,53 @@
...
+    .map(poetAnchorWithBlockData(block))
+
+function transactionToPoetAnchor(transaction: Transaction): TransactionPoetTimestamp | undefined {
```
warrenv: This can be written as a pipe() and would avoid the need for all the intermediate variables.
Any reason to use the function keyword vs an arrow function? It would keep it consistent with the other functions in the file.

package.json
```diff
@@ -40,6 +40,7 @@
   "dependencies": {
     "@po.et/poet-js": "2.0.0",
     "amqplib": "0.5.2",
+    "bitcoin-core": "^2.0.0",
```
kennylavender: remove "^"

README.md
```diff
@@ -121,6 +123,11 @@ This is what the default configuration looks like:
   mongodbUrl: 'mongodb://localhost:27017/poet',
   ipfsUrl: 'http://localhost:5001',
   insightUrl: 'https://test-insight.bitpay.com/api',
+  bitcoinUrl: '',
```
kennylavender: What is this bitcoinUrl, what is it and where do I get it? Maybe explain those things a bit more in the readme.

Same for the other configuration settings.

geoffturk: Don't we need to remove `insightUrl` now too?

src/BlockchainReader/Bitcoin.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import { blockToPoetAnchors } from './Bitcoin'
+
+import * as TestBlock from './TestData/block-00000000000151360aad32397ff1cf7dd303bed163b0ef425e71a53ccdec7312.json'
+
+describe('Bitcoin.blockToPoetAnchors', async should => {
+  const { assert } = should()
+
+  assert({
+    given: 'testnet block 00000000000151360aad32397ff1cf7dd303bed163b0ef425e71a53ccdec7312',
```
kennylavender: Little less specific here and a bit mored descriptive would be a sligh improvement I think, `given: a 'valid testnet block'`.

README.md
kennylavender: @geoffturk `insightUrl` is still used by the BlockchainWriter, this is just the Reader conversion.

src/BlockchainReader/Bitcoin.test.ts
```diff
@@ -0,0 +1,110 @@
+import { PoetTimestamp } from '@po.et/poet-js'
+import { allPass, equals } from 'ramda'
+import { describe } from 'riteway'
+
+import { PREFIX_POET, PREFIX_BARD } from 'Helpers/Bitcoin'
```
geoffturk: `PREFIX_POET` is not used anywhere in this file.

src/BlockchainReader/Bitcoin.ts
lautarodragan: In general I rather avoid adding null/undefined/empty input checking to functions so their scope and signature are more specific.

I'm thinking I could replace `.filter(_ => _)` with `.filter(identity)` (importing `identity` from Ramda), thought. What do you think?

src/BlockchainReader/Bitcoin.ts
lautarodragan: > Any reason to use the function keyword vs an arrow function? It would keep it consistent with the other functions in the file.

I guess I wrote it that way since it has a body and return statement, but either way works fine with me. I'll switch to arrow.

> This can be written as a pipe() and would avoid the need for all the intermediate variables.

I actually started writing it as a pipe originally, but I felt it didn't add as much value as not losing type information, which helps prevent bugs and code faster a great deal. The function is pure either way.

I do agree using pipe leaves less room for bugs (since it removes the need to have a function body and imperative statements completely). Once `pipe` is properly typed I'm definitely switching to it, but for now I feel it adds more value to have the IDE tell me something is wrong in real time for every key stroke and providing autocomplete/intellisense, and the build directly failing if I make a typo.

I'm hoping `pipe` will have proper types soon, I believe TS 3 should support it.

src/BlockchainReader/Bitcoin.test.ts
lautarodragan: I'm actually validating things that are only true for this particular block, though.

Do you still feel it's better to be less specific? (I'll change it if you do)

src/BlockchainReader/Bitcoin.test.ts
lautarodragan: @warrenv does [this commit](https://github.com/poetapp/node/pull/383/commits/8790c3a466fa46fdab35fe90923e97ecf56af1ee) fix it?

src/BlockchainReader/Bitcoin.ts
warrenv: I don't think using `identity` would improve it. If it's a pattern you use alot, I'd extract it to a utility and give it a more meaningful name, like `stripNulls` or something.

src/BlockchainReader/Bitcoin.test.ts
```diff
@@ -38,50 +38,52 @@ describe('Bitcoin.blockToPoetAnchors', async should => {
 
     const anchorsIpfsHashes = poetAnchors.map(poetAnchor => poetAnchor.ipfsDirectoryHash)
 
+    const given = 'a test block with Po.et anchors, blockToPoetAnchors'
```
warrenv: blockToPoetAnchors is already covered in the describe() so it seems redundant here.

lautarodragan:
@kennylavender yeah! 

Check this out:
- https://github.com/poetapp/node/pull/384
- https://github.com/poetapp/node/pull/385
- https://github.com/poetapp/node/pull/386
- https://github.com/poetapp/node/pull/387

package.json
lautarodragan: Addressed in separate PR

src/BlockchainReader/Bitcoin.ts
lautarodragan: Scratch that, `ìdentity` turns everything into `any` until we install `@types/ramda`, but that requires some changes since it causes the build to fail. I'll try to fix things and install `@types/ramda` in a separate PR after mainnet.

src/BlockchainReader/Bitcoin.test.ts
lautarodragan: fixed!

src/BlockchainReader/Bitcoin.test.ts
lautarodragan: fixed!

src/BlockchainReader/Bitcoin.test.ts
lautarodragan: Maybe if I change the hash for the height? (Less aggressive to the eye?)

README.md
lautarodragan: @kennylavender now that you mention it, you made me remember something.

This experience is rather painful and not in line with the other dependencies.

I should add a bitcoin-core running in testnet to the docker-compose file and change the default configuration so it works out-of-the-box with that docker image of bitcoind, as we're doing with IPFS, RabbitMQ and MongoDB.

I think I'll start working on that and then update the README to reflect so.

src/BlockchainReader/Bitcoin.ts
lautarodragan: What about [this](https://github.com/poetapp/node/pull/383/commits/de1699c1a18fd44d03cd6f456b71f3066f097d7e)?

src/BlockchainReader/Bitcoin.test.ts
lautarodragan: See [this commit](https://github.com/poetapp/node/pull/383/commits/242156343ab0031783ff369a517e1750144bf869)

src/BlockchainReader/Bitcoin.ts
warrenv: That's fine too.

src/BlockchainReader/Bitcoin.test.ts
kennylavender: Yup that helps make it less aggressive. Although, I would like given to explain what is special about this particular block if possible, `given a block that ...`, so that reading the test I know what is special about it.

lautarodragan:
@warrenv @kennylavender ready to review once again!

I've updated the PR's description now that its complexity has been reduced by pushing changes in separate PRs, and added info about the docker-compose.yml change I've pushed.

I've tested this both with a bitcoin-core running natively in AWS and running in my local machine in a docker container.

I've also updated the README further, @geoffturk can you review please?

geoffturk:
OK. I've created a separate PR for the README and assigned you as a reviewer.

src/BlockchainReader/ClaimController.ts
```diff
@@ -15,86 +17,77 @@ export class ClaimController {
...
 
-    if (matchingPoetTimestamps.length) await this.messaging.publishPoetTimestampsDownloaded(matchingPoetTimestamps)
+    if (matchingAnchors.length) await this.messaging.publishPoetTimestampsDownloaded(matchingAnchors)
   }
 
-  /**
-   * @returns {Promise<number>} The highest block height from the database, or null if no block has been processed.
-   */
-  async findHighestBlockHeight(): Promise<number> {
+  async findHighestBlockHeight(): Promise<number | null> {
```
wzalazar: I don't sure is good practice return a `null`

lautarodragan: Agreed. I personally don't see a problem with returning undefined or null as long as the signature makes it explicit, but in the future I'd like to try using an Option monad instead.

lautarodragan:
@kennylavender may not be Insight API this time — there's reports of the testnet being clogged up completely almost to the point of transactions not getting into blocks.

I'm using regtest to test the BlockchainWriter implementation — but for this PR, we don't really need to test broadcasting since that hasn't changed.

The best test, I think, is just to make sure the behavior hasn't changed if treating the node as a black box — so let the `master` node sync, see what works it downloads, let the `bitcoind-blockchain-reader` sync and confirm that it downloads the same stuff.

src/Helpers/isTruthy.ts
```diff
@@ -0,0 +1 @@
+export const isTruthy = (something: any): boolean => something
```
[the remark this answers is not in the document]
lautarodragan: My reasoning here is that I wanted to keep this function as simple as possible, and taking advantage of how JavaScript evaluates expressions, anything can be evaluated to `true` or `false`.

So even though it's implemented as an identity function, the TypeScript signature prevents the caller from actually knowing that it's returning the input — since it's signed as returning a `boolean`, TypeScript won't allow you to treat the result differently.

For example, `isTruthy({ theAnswer: 42 }).theAnswer` would fail to compile.

As such, I think it also doesn't need any unit tests or any extra logic (since it's only using a feature of the language). I think this is just normal use of JS truthy/falsey evaluation.

We could do this too:

```ts
import [ identity } from 'ramda'

export const isTruthy: (something: any) => boolean = identity

// or 

export const isTruthy = identity as (something: any) => boolean
```

If you fear there's room for bugs or it's unintuitive I'd go with the `!!` and add some unit tests.

What do you think?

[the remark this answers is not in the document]
kennylavender: I don't really like for the idea of using types to lie about function logic to provide logic. It's a very simple function so I almost didn't even bother making a comment, but I do see a possible issue if someone attempts to use this function and they pass in a value and expect a boolean back, it might take them a while to debug why they are getting non-boolean somewhere else down the line where they expected a boolean. I think we should go with `!!` and unit tests :)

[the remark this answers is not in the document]
lautarodragan: Done!

lautarodragan:
Merging this now. Let's raise GH issues if we find any bugs / code that could be improved.

PR #386, opened 2018-08-30
riteway typings

lautarodragan:

Since type declaration files are removed by tsc and don't exist at run time, the tests passing in Travis should be enough to indicate this PR is mergeable.


typings/riteway.d.ts
```diff
@@ -1 +1,16 @@
...
+
+  type describeCallback = (should: should) => Promise<void>
+
+  type should = (label?: string) => { assert: assert }
+
+  type assert = (assertion: Assertion) => void
+
+  interface Assertion {
+    given: string
+    should: string
```
kennylavender: `should` should be an optional property because will default to the label above in the `should` function.

lautarodragan: fixed. also added readonly, dunno if that works (is there any reason to ever mutate them?)

kennylavender: Nope, can't think of anything

Issue #93, opened 2018-05-07
Purify Code in BlockchainWriter


lautarodragan:
There's some logic in BlockchainWriter's `timestamp` that could be extracted. The method probably has too much responsibility, and some parts could be turned into pure functions, that could be unit tested.

https://github.com/poetapp/node/blob/77f0f3e861016a6fdf229f36c4a303d74dde9c05/src/BlockchainWriter/ClaimController.ts#L95-L100

https://github.com/poetapp/node/blob/77f0f3e861016a6fdf229f36c4a303d74dde9c05/src/BlockchainWriter/ClaimController.ts#L110-L114

Maybe there's more stuff?

Issue #293, opened 2018-08-02
Replace "timestamp" with a better term


kennylavender:
From @ericelliott 

https://github.com/poetapp/node/pull/142#issuecomment-402894564

> This is confusing to me. Claims should have their own timestamps before they get to the batching process. Batching to the blockchain doesn't timestamp the claims. It provides consensus for the order of claims, not the precise creation time of claims. I don't think we should be using "timestamping" in reference to claim batching.

> Also, batch frequency has no protocol-guaranteed timing, so if you think of a batch as a "timestamp", it doesn't have any particular resolution. Accuracy vary be within 20 minutes, an hour, or a week.

@ericelliott @lautarodragan @geoffturk To avoid confusion and miscommunication I was thinking we should use a better term throughout the codebase where we are currently using the term `timestamp`. I was thinking `anchor`? or maybe a more specific `blockchainAnchor`? This should also be adjusted in other repos such as poet-js. Thoughts?

lautarodragan:
Although it could cause confusion initially, I think it's pretty standard in the blockchain world. 

See [OpenTimestamps](https://opentimestamps.org/) for example.

The term may origin from [Trusted Timestamping](https://en.wikipedia.org/wiki/Trusted_timestamping). See [Decentralized timestamping on the blockchain](https://en.wikipedia.org/wiki/Trusted_timestamping#Decentralized_timestamping_on_the_blockchain).

I think the original concept arises from being able to prove that a certain document existed at a given time — _timestamping_ it onto a blockchain.

Searching for [timestamping to blockchain](https://www.google.com.ar/search?q=timestamping+to+blockchain) provides many results related to this.

geoffturk:
I hear what you are saying Lautaro, but just because it is a convention doesn't mean it shouldn't be improved if there is room for improvement. We could coin a new term like `blockchainAnchor` or `blockstamp` and then define it in our documentation.

I think it is a bit confusing when we use the same word to refer to two different things in the same system.

lautarodragan:
Agreed. If you guys still think it's confusing and would like to replace it, I'm all in for that.

geoffturk:
Great! `blockstamp` is short but maybe trying to be a bit too catchy. `blockchainAnchor` is good but a bit long. I'm fine with either one or maybe someone has a better idea.

What will this field contain exactly? The number of the block in the Bitcoin blockchain that the claim batch hash was written to?

kennylavender:
[...]

Issue #157, opened 2018-06-18
BlockchainWriter README


lautarodragan:
Add a README.md to the BlockchainWriter module.

See https://github.com/poetapp/node/pull/145#issuecomment-397540474 for an example, reasoning and discussion.

This issue can be tackled on its own or as part of any PR that introduces changes to the BlockchainWriter module.

Issue #27, opened 2017-12-05
Testing Environment Cross-Testing


lautarodragan:
The Testing Environment needs to have two completely isolated Po.et Nodes, each with its own IPFS, MongoDB and RabbitMQ. 

We need an integration test that sends an alert to Slack whenever a Work that has been timestamped by Node1 over an hour ago still hasn't been picked up by Node2.

warrenv:
This feels like it should be a smoke test versus an integration test.

PR #408, opened 2018-09-06
Replace Insight with Bitcoin Core in BlockchainWriter

lautarodragan:

Fixes https://github.com/poetapp/node/issues/24
Fixes https://github.com/poetapp/node/issues/93
Partly addresses https://github.com/poetapp/node/issues/293
needed-by https://github.com/poetapp/node/issues/22
needed-by https://github.com/poetapp/node/issues/25
needed-by https://github.com/poetapp/node/issues/27

### Release

## Description of Changes

This PR replaces the Insight API dependency of the BlockchainWriter module with the Bitcoin Core Wallet, finishing the migration.

### Docker Compose

- A service for regtest was added.
- The testnet service was renamed and its ports were fixed

### Dependencies

`bitcore-lib` and `@types/bitcore-lib` were uninstalled.

### Deleted Tests

Placeholder tests that were introduced to bypass an error in coverage were deleted.

### Rename

This PR started replacing `timestamp` with `anchor`. See https://github.com/poetapp/node/issues/293.

### Configuration

- The default configuration now points to regtest.
- `insightUrl`, `bitcoinAddress` and `bitcoinAddressPrivateKey` were removed.
- No new configuration entries were needed. The onces added in https://github.com/poetapp/node/pull/383 sufficed.

### Fee Estimation

The previous could would always send a fee of `0.001`. This new code is using the [`fundrawtransaction`](https://bitcoincore.org/en/doc/0.16.0/rpc/rawtransactions/fundrawtransaction/) RPC, which estimates fees based on recent blocks.

### How to Test

This is how I tested this change:

- `sudo docker-compose up -d bitcoin-regtest`
- `db.dropDatabase()`
- `npm run build`
- `npm start`
- `npm run test:integration`
- `GET http://localhost:18080/works`
- Verify that response contains a few items.
- After a configured amount of time these items should show `timestamp.ipfsFileHash`, `timestamp.ipfsDirectoryHash` and `transactionId`. These three properties will appear one by one. Keep refreshing until the show.
- `bitcoin-cli generate 1`
- Wait the configured amount of time until BlockchainReader picks up the generated block
- `GET http://localhost:18080/works` again
- The works in the response should now contain a full `timestamp`, with `blockHash`, `blockHeight`, etc.

I ran this test both in regtest and testnet. 

One ideal test I haven't done is node syncing: deploy this code and the dependencies to two different machines and confirm they sync correctly.

## Things Being Left Out

I'm leaving out of this PR a few things since it's actually working as it is and it's blocking several other PRs.

- Update documentation
- https://github.com/poetapp/node/issues/157
- Unit tests
- Possible refactors or code improvements


src/BlockchainWriter/ClaimController.ts
```diff
@@ -48,90 +45,89 @@ export class ClaimController {
...
+    await collection.updateOne({ ipfsDirectoryHash }, { $set: { txId } }, { upsert: true })
+    await messaging.publish(Exchange.IPFSHashTxId, {
+      ipfsDirectoryHash,
+      txId,
+    })
+  }
 
-    if (!utxo || !utxo.length)
-      throw new Error(`Wallet seems to be empty. Check funds for ${this.configuration.bitcoinAddress}`)
+  private anchorData = async (data: string) => {
```
lautarodragan: I experimented with `pipeP` for this but decided not to use it yet.

I've left the experiments in [another branch](https://github.com/poetapp/node/compare/bitcoin-core-writer...blockchain-writer-functional-experiments?expand=1).

src/BlockchainWriter/Bitcoin.ts
```diff
@@ -0,0 +1,2 @@
+export const getData = (prefix: string, version: ReadonlyArray<number>) => (message: string) =>
```
kennylavender: Can/Should this be unit tested?

src/BlockchainWriter/ClaimController.ts
kennylavender: Agreed, it probably requires to much refactoring to use it nicely atm.

lautarodragan:
Will add proper docs in a new PR.

For now:  in regtest, `generate` should give you funds. Remember to generate at least 101 blocks since mined/coinbase bitcoins cannot be used until they have 100 confirmations.

docker-compose.yml
```diff
@@ -41,21 +41,30 @@ services:
         sleep 10;
         npm start;
       "
-  poet-bitcoind:
+  bitcoin-testnet:
```
geoffturk: I don't see the need to stand up a testnet docker instance anymore now that we are using regtest. If that is correct, can we either comment this out or remove it? Otherwise when I run `docker up --build` I am starting up a testnet instance and syncing with the testnet blockchain, thus using up bandwidth and storage for no real reason.

kennylavender: Good point @geoffturk. I can't think of a situation that I would still want a testnet docker instance locally.

kennylavender: Actually, now that I thought about it more, I think we do want a mainnet/testnet instance actually for people to run the node on their machines. Maybe we just need a simpler way of expressing which instance we run instead of running all of them, but that can be tackled in a future PR.

lautarodragan: That's correct, I'll split it apart in different docker-compose files. We still want to test stuff in testnet if possible, but `docker up` shouldn't cause you to download the whole testnet.

docker-compose.yml
```diff
@@ -36,27 +36,20 @@ services:
       - poet-mongo
       - poet-rabbit
       - poet-ipfs
+      - bitcoin
```
jwicks31: should this be bitcoin-regtest?

lautarodragan: yes! good catch.

lautarodragan: fixed!

lautarodragan:
@kennylavender @geoffturk @jwicks31 merging now not to block other PRs. I think I've addressed your reviews, but please reach out / raise issues if there's changes you think should be done to this PR.

PR #477, opened 2018-09-23
chore: lint tests

### LINKED ISSUE #280: Unit tests excluded from TSLint
TSLint ignores files excluded in tsconfig.json.

We currently have the following configuration:

```json
  "exclude": [
    "./src/**/*.test.ts"
  ]
```

I can't find any way to indicate TSLint that we DO want these files to be linted. I think the best approach right now would be to simply not exclude them.

We currently have about 14 lint errors in `*.test.ts` files

warrenv:

### Release

## Description of Changes

Fixes #280

Previously, unit tests were not being linted due to an excludes config
in `tsconfig.json`. These rules were added so that tests did not end up in
the `dist/` directory after a build.

This commit adds an additional tsconfig file that removes the excludes.
This config file is passed when running the linting scripts.

It also disables "no-relative-import" since we are disabling it in individual files all over the repo anyway and many test files got flagged for this.


tsconfig.with-tests.json
```diff
@@ -0,0 +1,21 @@
+{
```
wzalazar: What do you say, if we keep all configuration on the same `tsconfig.json`. We can exclude the files in the build with `.npmignore` or `.dockerignore`

warrenv: That could work, but there may be instances where we want the test files to end up in a container (for example) for testing environment purposes. I chose this solution because it addresses the problem at it's root: typescript. If we push the problem lower down, we need to maintain multiple config files.

lautarodragan: I'd strongly rather not have two tsconfig.json files. It's very easy for them to fall out of sync.

Is there any issue with tests landing in dist, or is it just a cosmetic thing?

warrenv:
It would need to disabled in all of the test files if we don't disable by default. I did not want to experience that pain.

I feel we should address this relative-path issue separately as it is a problem across all repos.

tsconfig.with-tests.json
warrenv: This was the best recommendation I found after researching it.

Tests don't belong in the built container - they add bulk and could potentially open security issues.

lautarodragan:
@warrenv you're right. For now let's disable it manually for every test file, should not take longer than 2 minutes. Let's also create an issue in our tsling-rules repo so we can enhance the rule as required.

tsconfig.with-tests.json
lautarodragan: If they don't belong in the container, why not `.dockerignore` and `.npmignore` them as Walter suggested?

I get the feeling that, if it's only a container issue, it should be handled by the container tooling...

warrenv: As I mentioned, sometimes we do want tests in the container. By 'built container', I mean production container.

I agree this is not the most elegant solution by having two configs, but I believe it is the quickest way to move forward.

lautarodragan: Makes sense. Let’s keep it as is for now then

Issue #479, opened 2018-09-24
Extract message queueing config to env vars


warrenv:
Identify and extract rabbitmq related config (exchange names, etc) into `src/Configure.ts`.
Update necessary modules to use the config.

PR #480, opened 2018-09-24
Config rabbitmq

wzalazar:

### Release

## Description of Changes

Fix https://github.com/poetapp/node/issues/479
FIx https://github.com/poetapp/node/issues/481

This PR introduces:

- config exchange messages RabbitMQ
- unit test
- Replaces the `enum Exchange` to a hash table with the configuration


warrenv:
Nice work @wzalazar!

wzalazar:
@warrenv I appreciate so much your help :sunglasses:

src/Messaging/Messages.ts
```diff
@@ -1,16 +1,19 @@
-export enum Exchange {
-  BatchReaderReadNextDirectoryRequest = 'BATCH_READER::READ_NEXT_DIRECTORY_REQUEST',
-  BatchReaderReadNextDirectorySuccess = 'BATCH_READER::READ_NEXT_DIRECTORY_SUCCESS',
-  BatchWriterCreateNextBatchRequest = 'BATCH_WRITER::CREATE_NEXT_BATCH_REQUEST',
-  BatchWriterCreateNextBatchSuccess = 'BATCH_WRITER::CREATE_NEXT_BATCH_SUCCESS',
+import { loadConfigurationWithDefaults } from 'Configuration'
+const configuration = loadConfigurationWithDefaults()
```
lautarodragan: `loadConfigurationWithDefaults` does IO. It shouldn't be called in global scope.

You'll actually need to pass the needed configuration entries to each module from app.ts and have each module inject them so the router knows about them.

It's a pretty easy but boilerplate-y change. That's the main reason I wanted to avoid making every exchange configurable.

I'm OK with merging this pr-as is if it's blocking something else, but the refactor should start immediately after the merge.

warrenv: Sorry, but why is I/O in global scope bad?

lautarodragan: Two mains reasons:
- The most immediate one is that `import { app }` will trigger these calls. Side effects running on an import are generally unexpected and can lead to hard-to-find bugs
- You have no control over statements running in global scope. You can't import them, integration-test them, etc. JS doesn't support `lazy` so it also becomes hard to understand how the code flows.

It's usually best to have the whole application being an exported thing that does nothing by itself and a file that imports and runs it, as we're doing with `index.ts` and `app.ts`.

geoffturk: > I'm OK with merging this pr-as is if it's blocking something else, but the refactor should start immediately after the merge.

This one is blocking other things from happening today, so I suggest merge it in and then create a ticket to refactor.

PR #484, opened 2018-09-25
refactor: allow tests to inject config

### LINKED ISSUE #483: inject config into app()
To avoid mutating process.env in tests, accept a config object as input for the app() function now that it has been separated out into it's own module.

warrenv:

### Release

## Description of Changes

Fixes #483 

Refactor the app module to take a config object that overrides all default and environment config.
This is needed for configuration setup in integration and functional tests so we don't
have to mutate process.env in our tests.


src/Configuration.test.ts
```diff
@@ -81,3 +81,20 @@ describe('src/Configuration', async (should: any) => {
     }
   }
 })
+
+describe('loadConfigurationWithDefaults', async (should: any) => {
```
lautarodragan: Since `loadConfigurationWithDefaults`, isn't this more an integration test than a unit test?

Issue #497, opened 2018-09-28
Fix docker-compose setup


warrenv:
We need to rollback https://github.com/poetapp/node/pull/485 because docker-compose.override.yml was checked in and this should never be the case. 

This is my fault as I did not pay enough attention to the filenames during my PR review.

We need to have a video meeting to discuss what issue this was trying to solve and come up with a solution that meshes with the way docker-compose is supposed to be used.

warrenv:
Here are some useful blog posts describing how docker-compose can be used for handling multiple environments including development.

https://runnable.com/docker/advanced-docker-compose-configuration
https://medium.com/@basi/docker-compose-from-development-to-production-88000124a57c
https://pspdfkit.com/blog/2018/how-to-manage-multiple-system-configurations-using-docker-compose/

PR #499, opened 2018-10-01
chore: refactor docker-compose files

warrenv:

### Release

## Description of Changes

Fixes: #497 

This refactors the docker compose config files to reverse
an earlier commit that introduced `docker-compose.override.yml` into
the repo. This file is used to alter the service startup for customizing local development
needs so it should never be checked in.

- All services have been moved back into `docker-compose.yml`
- A unit test for load configuration was made more robust so that
  using environment variable overrides do not break it.
- bitcoind services were renamed to be more generic and which
  network they connect to was made configurable for better reuse.


docker-compose.override.yml.example
```diff
@@ -1,12 +1,40 @@
 version: '3'
 
 services:
+  rabbitmq:
```
geoffturk: Shouldn't it just be called `rabbit` like in `docker-compose.yml` below?

warrenv: yes.

warrenv:
I was going to leave testnet for a future refactor since I haven't dug into where it's being used yet.
Using env vars to choose test or regtest in this refactor is a move to potetially eliminating the testnet docker-compose.

docker-compose.yml
```diff
@@ -8,22 +45,29 @@ services:
...
-      - BITCOIN_URL=bitcoin-regtest
-      - RABBITMQ_URL=amqp://poet-rabbit
-      - MONGODB_URL=mongodb://poet-mongo:27017/poet
-      - IPFS_URL=http://poet-ipfs:5001
+      - POET_SERVICE=node
+      - BITCOIN_URL=bitcoind-1
+      - RABBITMQ_URL=amqp://rabbit
+      - MONGODB_HOST=mongo
+      - IPFS_URL=http://ipfs:5001
       - INTEGRATION_TEST_MONGODB_URL=mongodb://poet-mongo:27017/poet
```
wzalazar: Should be `mongodb://mongo:27017/poet` ?

warrenv: It's not actually being used anywhere - I will remove them.

PR #522, opened 2018-10-05
refactor: secure mongo and add database test helpers

### LINKED ISSUE #521: Secure mongo
The current docker setup runs mongo without any auth. In order to provide better security, copy the config from local-dev and update the functional tests so they do not break.

warrenv:

### Release

## Description of Changes

Closes #521

Previously, mongo was being run without any auth.

This commit:
  - Enables mongo auth for a better security profile.
  - Adds database helpers to create and teardown temporary databases
     during testing.
  - Updates the functional tests to use the new database helpers.

To test:
```bash
  $ cp docker-compose.override.yml.example docker-compose.override.yml
  $ docker-compose build
  $ docker-compose up -d
  $ docker-compose exec poet-node test:functional
```


jwicks31:
I pulled this branch down to review [docker-compose-refresh](https://github.com/poetapp/local-dev/pull/53) and using both the local-dev environment and the node docker-compose file I get 8 failing tests out of 12.

warrenv:
Are you seeing any obvious errors in the output?

jwicks31:
I re-ran the tests inside of both environments. It passed for all 12 this time in the node environment, but when using local dev the same 8 still failed. The reason they are failing is because: 
```
operator: deepEqual
expected: 'string'
actual:   'undefined'
```

They are checking for the `blockHeight`, `blockHash`, `transactionId` in Node A. And the `ipfsFileHash`, `ipfsDirectoryHash`, `blockHeight`, `blockHash`, `transactionId` in Node B.

docker-compose.override.yml.example
```diff
@@ -14,23 +17,23 @@ services:
   # running functional tests. Someone checking out the repo and wanting
   # to just run a poet node only needs bitcoind-1.
   bitcoind-2:
-    # image: ruimarinho/bitcoin-core:0.16.3
+    image: ruimarinho/bitcoin-core:0.16.3
```
geoffturk: 0.17.0

mongodb/create_poet_db.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
lautarodragan: Will this script be only used in tests, or also in deployments for production?

warrenv: I'm guessing production would be different - I'm not 100% clear on how production is set up (are we using Atlas?).

geoffturk: Yes. We are using Atlas.

Ping @WesleyCharlesBlake regarding the use of the script (in production and QA).

WesleyCharlesBlake: this script will not be used on the Atlas mongodb clusters. Atlas allows us to setup users, and specify perms on DBs and collections through its UI, and the DB's get created when we specify the connection URI in the mongo client (ie `mongodb+srv://admin:<PASSWORD>@<some-cluster-id>.mongodb.net/test` will end up creating a `test` db, provided that the credentials provided in the URI have permissions to read/write on that DB.

from a K8s perspectice, we supply the mongo Atlas DB creds in values.yaml, and this creates a config secret in K8s which the node deployment then retrieves to configure mongo auth

docker-compose.override.yml.example
```diff
@@ -14,23 +17,23 @@ services:
...
-    #   -deprecatedrpc=signrawtransaction
-    #   -addnode=bitcoind-1
-    # links:
-    #   - bitcoind-1:bitcoind-1
+    ports:
+      - "19443:18443"
+    command:
+      -printtoconsole
+      -testnet=${USE_BTC_TESTNET:-0}
+     -regtest=${USE_BTC_REGTEST:-1}
```
krobi64: add space

krobi64:
I also can not run `docker exec poet-node test:functional`. It says `Error: No such container: poet-node`.

However I can use `docker-compose exec poet-node bash` and then `npm run test:functional`.

That returns 
```
1..12
# tests 12
# pass  5
# fail  7
```

mongodb/create_poet_db.sh
lautarodragan: OK. Thanks @geoffturk and @WesleyCharlesBlake.

We should move this file to /tests then. /scripts could work too, if we intend to run it manually.

warrenv: I disagree - this is not for tests, it is for bootstrapping a brand new instance of the mongo docker container. It's a hook into how the mongo containers work so you can have an existing set of databases and users already set up when the container starts.

docker-compose.override.yml.example
warrenv: done.

warrenv:
@krobi64 

- I updated the instructions to use docker-compose instead of docker.
- The 'out of range errors' are normal - they do not affect the tests.

Sometimes the functional tests fail if they are run immediately after the container start.
Re-running them usually gives a passing result. If not. , it may be necessary to bump the timeout values in the poet-node section of your docker-compose.override.yml:

TIMESTAMP_INTERVAL_IN_SECONDS
BATCH_CREATION_INTERVAL_IN_SECONDS
READ_DIRECTORY_INTERVAL_IN_SECONDS

krobi64:
* [x] Tested changes manually
* [x] Checked accidental architectural/style changes
* [x] Reviewed entire diff
* [ ] Unit tests
* [ ] Documentation
* [x] Filenames and locations

Issue #46, opened 2018-02-21
BlockchainWriter Confirmations


lautarodragan:
Add a check to BlockchainWriter that makes sure the broadcasted transaction has at least `N` (configurable, default to `1`) confirmations.

It should emit a new type of RMQ message, too: `BlockchainTransactionConfirmed`.

If too much time has passed and the transaction hasn't been confirmed, discard it and let a new transaction be created and broadcasted.

geoffturk:
needs #157

PR #524, opened 2018-10-06
Health Endpoint

### LINKED ISSUE #376: Health Endpoint
Should return info about
- connectivity to IPFS, RMQ, DB, Bitcoin Core
- state of retries
- state of ipfs hashes that were given up
- failed transaction broadcasts
- bitcoin core getblockchaininfo 
- - blocks
- - verificationprogress
- - bestblockhash
- - warnings
- - size_on_disk
- bitcoin core getnetworkinfo 
- - version
- - subversion
- - connections
- - networkactive
- - protocolversion
- - warnings
- bitcoin core getwalletinfo 
- - balance
- - txcount
- TBD

### Light Mode

Should support a `light` mode that does no heavy operations so it can be queried on a short interval and a `heavy` mode that offers greater insight.

### Implementation

Probably we'll want to add a new Health module responsible for collecting and aggregating this information.

The new endpoint will need to be added to the API module, so we'll have to figure out the best way for the API module to read this information. Most likely just let the API module directly read from the Health Module's DB, same way we're doing with the View module.

jwicks31:

### Release

#### Health Module
Added a Health Module with the following responsibilities:

##### Health
Entry Point for the Module. Initializes the container, starts the Router, and HealthService.

##### Health Service
At a set interval checks MongoDb Connection, IPFS Connection, and gathers Bitcoin-Core information. Storing each in the database.

##### Router
Consumes `ClaimsNotDownloadedMessages and stores IFPS information in the database. Information Stored is: `ipfsFileHash`, `failureReason`, `failureType`, `downloadAttempts`, and `lastDownloadAttemptTime`. 

##### IPFS
API for IPFS service.

##### HealthController
Supplies a getHealth method that queries the mongoDB and returns the Health information:
- connectivity to IPFS, DB, and Bitcoin Core
- state of retries
  - ipfsDownloadFailures(array of all failures with reason, type, downloadAttempts, lastDownloadAttemptTime, and ipfsFileHash.)
  - hardIpfsFailures(array of all hard failures - same info as ipfsDownloadFailures)
  - ipfsFailureWillRetry(array of soft failures that have retry attempts left - same info as ipfsDownloadFailures)
  - ipfsFailureMaxRetries(array of soft failures that have reached max retry attempts - same info as ipfsDownloadFailures)
- state of ipfs hashes that were given up
- bitcoin core getblockchaininfo 
- - blocks
- - verificationprogress
- - bestblockhash
- - warnings
- - size_on_disk
- bitcoin core getnetworkinfo 
- - version
- - subversion
- - connections
- - networkactive
- - protocolversion
- - warnings
- bitcoin core getwalletinfo 
- - balance
- - txcount

#### TODO
- failed transaction broadcasts
- connectivity to RMQ.

#### Notes
I haven't implemented the `light` or `heavy` modes yet as I wasn't sure which would fall in to which category. This is a first attempt at the endpoint and I'm not sure that I took the best route in gathering the information. 

My initial thought was that we would have a Health Service with a `light` and `heavy` mode that gathered information periodically that was always available to query, but I think that I am not realizing that it might have meant 2 different endpoints with one collecting the `light` info and one gathering the further insight. 

I also wasn't sure how to check the connectivity to RMQ or what information we are wanting to gather for the failed transaction broadcasts.

Resolves #376


warrenv:
I don't think `package-lock.json` or `wait-for-it.sh` should be part of this change.
There are no changes to `package.json` and `wait-for-it.sh` is a side-effect of mounting the project in a container during a docker-compose build.

src/API/Router.ts
```diff
@@ -90,6 +95,31 @@ export class Router {
...
+      mongoIsConnected,
+      ipfsIsConnected,
+      blockchainInfo,
+      walletInfo,
+      networkInfo,
+      ipfsDownloadRetries,
+      hardIpfsFailures,
+      ipfsFailureWillRetry,
+      ipfsFailureMaxRetries,
+    } = await this.healthController.getHealth()
```
lautarodragan: Isn't is simpler to just `context.body = await this.healthController.getHealth()`?

lautarodragan:
@jwicks31 this is amazing.

> I also wasn't sure how to check the connectivity to RMQ or what information we are wanting to gather for the failed transaction broadcasts.

For failed transaction broadcasts we'll need https://github.com/poetapp/node/issues/46 first, so it's okay to leave out of this PR.

Testing RMQ can be tricky. I'm thinking we could have some sort of ping RMQ message and track big delays in it. ampqlib will usually throw an error if something goes awry though, so that'll end up in our logs. We can leave more solid RMQ testing out of this PR.

src/Health/Router.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.messaging = messaging
+    this.controller = controller
+  }
+
+  async start() {
+    await this.messaging.consumeClaimsNotDownloaded(this.onClaimsNotDownloaded)
+  }
+
+  async stop() {
+    this.logger.info('Stopping Health Router...')
```
lautarodragan: I think this line may be a left over?

src/Health/HealthController.ts
```diff
@@ -0,0 +1,205 @@
...
+    this.db = db
+    this.collection = this.db.collection('health')
+    this.bitcoinCore = bitcoinCore
+    this.ipfs = ipfs
+  }
+
+  async checkMongoConnection(): Promise<any> {
+    this.logger.trace('MongoHealth', 'Checking Mongo...')
+    const connection = await this.db.stats()
+    const mongoIsConnected = connection.ok === 1 ? true : false
```
lautarodragan: You can just do `const mongoIsConnected = connection.ok === 1` here, leaving ` ? true : false` out has the same effect

src/Health/HealthController.ts
```diff
@@ -0,0 +1,205 @@
...
+        $set: {
+          mongoIsConnected,
+        },
+      },
+      { upsert: true }
+    )
+  }
+
+  async checkIpfsConnection(): Promise<any> {
+    this.logger.trace('ipfsHealth', 'Checking IPFS...')
```
lautarodragan: You can leave the first argument out here, changing it to `this.logger.trace('Checking IPFS...')`.

This will log the following:

```
[2018-10-10T00:31:07.250Z] TRACE (31451 on Laumint): Checking IPFS...
    module: "Health"
    file: "HealthController"
```

The `module` and `file` properties that `this.logger` already provide enough info in this case IMO.

If you'd like to add the method name like we're doing in other places in the code base, please pass an object as the first parameter, so it gets merged (instead of the string getting concatenated), same way you're doing it in `getBlockchainInfo`.

src/Health/HealthController.ts
```diff
@@ -0,0 +1,205 @@
...
+  async getBlockchainInfo(): Promise<void> {
+    const logger = this.logger.child({ method: 'getBlockchainInfo' })
+
+    const {
+      blocks,
+      verificationprogress,
+      bestblockhash,
+      warnings,
+      size_on_disk,
+    } = await this.bitcoinCore.getBlockchainInfo().then((res: any) => res)
```
lautarodragan: `.then((res: any) => res)` is redundant, it can be removed safely.

lautarodragan:
@jwicks31 disregard my original comments on the `light` and `heavy` modes, I think I mixed my ideas in there. Your implementation is already _light_ (no heavy processing upon endpoint request) but has all the data in it.

src/API/HealthController.ts
```diff
@@ -0,0 +1,82 @@
+import { inject, injectable } from 'inversify'
+import { Db, Collection } from 'mongodb'
+import { HealthControllerConfiguration } from './HealthControllerConfiguration'
```
kennylavender: Line break before and after this line.

src/Health/HealthController.ts
```diff
@@ -0,0 +1,205 @@
...
+    this.collection = this.db.collection('health')
+    this.bitcoinCore = bitcoinCore
+    this.ipfs = ipfs
+  }
+
+  async checkMongoConnection(): Promise<any> {
+    this.logger.trace('MongoHealth', 'Checking Mongo...')
+    const connection = await this.db.stats()
+    const mongoIsConnected = connection.ok === 1 ? true : false
+    await this.collection.updateOne(
```
kennylavender: If mongo is not connected, it will never set this to false in mongo. So this work of storing the mongo connection status in the db is not helpful. Can you think of a better way?

src/API/HealthController.ts
```diff
@@ -0,0 +1,82 @@
...
+  constructor(
+    @inject('DB') db: Db,
+    @inject('HealthControllerConfiguration') configuration: HealthControllerConfiguration
+  ) {
+    this.db = db
+    this.collection = this.db.collection('health')
+    this.configuration = configuration
+  }
+
+  async getHealth(): Promise<any> {
```
kennylavender: We are relying on mongo being up here in order to get any health information. IMO it shouldn't be required for mongo to be working to get a health status of connections to services. services being mongo, ipfs, rabbitmq, blockchain info, wallet info etc.

We will need mongo for health information about specific failures and stuff, but that is extra information.

@lautarodragan not sure if you have anything to add. disagree/agree.

kennylavender: Also, I would split up this PR. Tackling everything requested in this issue all in one PR is not necessary.

src/Health/HealthController.ts
lautarodragan: You're right. If the connection to mongo is broken, we should get error log messages that will trigger alarms, though. Maybe just [confirm this and] remove the information about mongo?

It's the same thing for RMQ, and we should probably follow the same logic for IPFS and Bitcoin Core, though we'd probably need to update the code to check _what_ error is being thrown, due to the retry systems.

src/API/HealthController.ts
lautarodragan: See my response to your other comment on this matter.

I'm thinking now: if MDB is down, there's no way to persist state, so the application should crash. 

Regarding splitting this PR: I'd like your input @kennylavender on how information about retries is being stored in the Health module. @jwicks31 you're storing information inside a single object in the collection, in an array, right? Haven't been able to test it and see how the collection looks in that case yet, but it's probably better to just store aggregated data (just counts, no information about individual IPFS hashes). 

The results of calling `GET /health` should indicate whether there's a problem and trigger alarms, but need not provide debug detail on the problem. In the case Health signals a problem, we'll want to clone production to another environment and manually debug it, most likely. I believe we can have automated replicas without much effort with Atlas (@WesleyCharlesBlake can you confirm that?) 

If we choose to go that path, we could do it in a separate PR and simplify this one for now.

kennylavender: @lautarodragan 

> but it's probably better to just store aggregated data (just counts, no information about individual IPFS hashes)

yup, I agree, just counts, no details, just enough info to alert us to an issue to look into.

## Functional Tests
I think we should have functional tests for this endpoint too so we don't accidentally break these health checks. This would probably require collab with @warrenv . This is another good reason to split this PR and focus on getting the functional tests in place as well.

lautarodragan: It's decided then. Let's do just counts, no details @jwicks31.

Some basic functional tests should be simple enough to set up if you wanna give it a go: 

```ts
// create work etc
await delay(n)
const health = client.getHealth()
// assert predictable changes to health.blocks, etc
```

kennylavender:
@lautarodragan 

> Haven't tested the retry cases yet, which have some complex queries in them. I'll try to test them. @kennylavender any suggestions on how to do that?

**Manually testing**  
manually testing this would be a pain, the only thing I can think of its modifying the code to purposely throw the issues.

**Functional tests**
Don't think these fit as we need to purposefully create issues that are not possible to create from the API.

**Integration tests**
maybe we can come up with some integration tests that cover these cases

lautarodragan:
> Manually testing them should be enough. But we do need at least some very basic manual testing to ensure the health endpoint is working.

Which is another valid reason to split the PR - makes reviewing and testing easier. Part of this PR is easy to test.

jwicks31:
Closing this PR. I'm going to open up smaller PR's that introduce less changes.

src/API/HealthController.ts
jwicks31: @lautarodragan Yes this is all currently stored inside a single object in the collection, and it is an Array. I'll update it so that we aren't storing any additional information, just the counts of IPFSFailures, IPFSHardFailures, IPFSSoftWIllRetry, and IPFSSoftWontRetry. 

As far as collecting the information from Bitcoin-Core - should we collect this information on an interval and sore the information in a database to be retrieved whenever the /health route is pinged? Or should we just get that information as the /health route is pinged and not store it?

PR #539, opened 2018-10-10
Health Controller + IPFS for API Module

jwicks31:

### Release

## Description of Changes

Part of #376

##### IPFS
API for IPFS service.

##### HealthController
Supplies a getHealth method returns the Health information:
- connectivity to IPFS and MongoDB


src/API/IPFSConfiguration.ts
```diff
@@ -0,0 +1,5 @@
+/* tslint:disable:interface-name */
+
+export interface IPFSConfiguration {
```
kennylavender: Please move this interface into the class file because it belongs to the class. We are no longer separating class configuration interfaces from the class.

src/API/Router.test.ts
```diff
@@ -4,6 +4,9 @@ import * as Pino from 'pino'
 import { describe } from 'riteway'
```
lautarodragan: This is unrelated to your PR, but can you just delete this file?

We introduced a bunch of fake tests just to hack our way around coverage a while back and decided it was a bad idea quickly afterwards. 

I keep seeing PRs having to deal with these fake tests to keep the code passing. Better to just delete them IMO.

@kennylavender FYI

src/API/HealthController.ts
```diff
@@ -0,0 +1,28 @@
...
+  private readonly db: Db
+  private readonly ipfs: IPFS
+
+  constructor(@inject('DB') db: Db, @inject('IPFS') ipfs: IPFS) {
+    this.db = db
+    this.ipfs = ipfs
+  }
+
+  async getHealth(): Promise<any> {
+    const mongoConnection = await this.db.stats()
```
kennylavender: Each function should do 1 thing. You can move the mongo and ipfs logic into their own private methods. getHealth should only be composing together the results of its own private methods, it shouldn't know about external stuff.

```
async getHealth(): Promise<any> {
  const mongoStatus = this.getMongoStatus()
  const ipfsStatus = this.getIPFSStatus()
  // ...stuff
}
```

src/API/HealthController.ts
```diff
@@ -0,0 +1,28 @@
...
+    const mongoIsConnected = mongoConnection.ok === 1
```
kennylavender: make this logic `mongoConnection.ok === 1` a function and unit test it. The function can be external of the class so that we don't' need to instantiate the class.

src/API/HealthController.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const ipfsConnection = await this.ipfs.checkHealth()
+    const ipfsIsConnected = ipfsConnection === 200
```
kennylavender: create a function and unit test this logic as well.

src/API/IPFS.ts
```diff
@@ -0,0 +1,22 @@
...
+
+@injectable()
+export class IPFS {
+  private readonly url: string
+
+  constructor(@inject('IPFSConfiguration') configuration: IPFSConfiguration) {
+    this.url = configuration.ipfsUrl
+  }
+  
+  checkHealth: checkHealth = async () => {
```
lautarodragan: It's generally better for application clients to match the API one-to-one, and have our business logic in another layer (controller usually).

IOW, exposing a `getVersion` here, and doing the ipfsIsConnected logic in the controller.

Since you're relying on the status, which shouldn't leave the IPFS client, maybe `throw` if `status !== 200`. I believe `fetch` already throws for network errors, so if mongo is unavailable even to respond with a `500`, we'll get an error thrown anyway.

src/API/HealthController.ts
lautarodragan: does `db.stats()` never throw? if the connection to the db is broken completely, does it return an object anyway?

src/API/HealthController.ts
kennylavender: This might go out the window with @lautarodragan 's suggestion :)

src/API/HealthController.ts
kennylavender: good question, will leave to you @jwicks31

jwicks31: The [Mongo Docs](https://docs.mongodb.com/manual/reference/method/db.stats/#accuracy-after-unexpected-shutdown) make it seem to me like it will return an object. 

I've tried testing manually by giving incorrect ENV variables for Mongo, but the app crashes. So I wasn't able to verify what is returned.

 I also tested by closing the MongoClient right after starting it inside of the API module. This did throw.

src/API/IPFS.ts
jwicks31: Ok I changed this up. I'm not sure I understood what you mean in the last paragraph though.

jwicks31:
@kennylavender Just committed the functional tests. 

Thank you both for reviewing!

src/API/HealthController.ts
```diff
@@ -0,0 +1,45 @@
...
+  async checkMongo(): Promise<any> {
+    try {
+      const mongoConnection = await this.db.stats()
+      return isOkOne(mongoConnection)
+    } catch (e) {
+      return false
+    }
+  }
+
+  async checkIPFS(): Promise<any> {
```
kennylavender: make this private, readonly and it's returning a promise of boolean.

kennylavender: make this private and it's returning a promise of boolean.

src/API/HealthController.ts
```diff
@@ -0,0 +1,45 @@
...
+export class HealthController {
+  private readonly db: Db
+  private readonly ipfs: IPFS
+
+  constructor(@inject('DB') db: Db, @inject('IPFS') ipfs: IPFS) {
+    this.db = db
+    this.ipfs = ipfs
+  }
+
+  async checkMongo(): Promise<any> {
```
kennylavender: make this private and it's returning a promise of boolean.

src/API/HealthController.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  async checkIPFS(): Promise<any> {
+    try {
+      const ipfsConnection = await this.ipfs.getVersion()
+      return isStatus200(ipfsConnection)
+    } catch (e) {
+      return false
+    }
+  }
+  async getHealth(): Promise<any> {
```
kennylavender: line break above

src/API/HealthController.ts
lautarodragan: Thanks @jwicks31! App crashing or throwing works. Errors thrown in API get caught in a middleware and logged, so alarms will get triggered.

tests/functional/health.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const blockchainSettings = {
+  MINIMUM_BLOCK_HEIGHT: 100,
+  ENABLE_TIMESTAMPING: true,
+  TIMESTAMP_INTERVAL_IN_SECONDS: 10,
+  BATCH_CREATION_INTERVAL_IN_SECONDS: 5,
+  READ_DIRECTORY_INTERVAL_IN_SECONDS: 5,
+}
+
+const createDatabase = async (prefix: string) => {
```
lautarodragan: Isn't this the same as `claim_with_data.ts`'s `createDatabase`? If so, can we move the function to a helper instead?

tests/functional/health.ts
```diff
@@ -0,0 +1,57 @@
...
+    // Check health.
+
+    const response = await getHealth(NODE_PORT)
+    const healthData = await response.json()
+
+    const { mongoIsConnected, ipfsIsConnected } = healthData
+
+    assert({
+      given: 'a request to the health endpoing',
+      should: 'return mongoIsConnected and ipfsIsConnected',
```
warrenv: remove 'and ipfsIsConnected'?

tests/functional/health.ts
```diff
@@ -0,0 +1,57 @@
...
+      actual: typeof mongoIsConnected,
+      expected: 'boolean',
+    })
+
+    assert({
+      given: 'a request to the health endpoing',
+      should: 'return mongoIsConnected and ipfsIsConnected',
```
warrenv: remove 'and mongoIsConnected'?

tests/functional/health.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('Health Endpoint Returns the Correct Fields', async (assert: any) => {
+  const db = await createDatabase(PREFIX)
+  const server = await app({
+    BITCOIN_URL: process.env.BITCOIN_URL || 'bitcoind-1',
+    API_PORT: NODE_PORT,
+    MONGODB_DATABASE: db.settings.tempDbName,
+    MONGODB_USER: db.settings.tempDbUser,
+    MONGODB_PASSWORD: db.settings.tempDbPassword,
+    EXCHANGE_PREFIX: PREFIX,
+    ...blockchainSettings,
```
warrenv: `blockchainSettings` can be removed here and above. The default values are ok since this test is not doing anything to checking claims, etc. so timing is not an issue.

src/app.ts
```diff
@@ -49,6 +49,7 @@ export async function app(localVars: any = {}) {
     port: configuration.apiPort,
     dbUrl: configuration.mongodbUrl,
     rabbitmqUrl: configuration.rabbitmqUrl,
+    ipfsUrl: configuration.ipfsUrl,
```
lautarodragan: API shouldn't need to use IPFS - it's the HealthModule's responsibility to collect health data

jwicks31: Ok. So we would want the health module to check for IPFS connectivity and then store it in mongo to be read by the API module?

lautarodragan: Yep, your initial approach was right about this. 

And yeah, no problem doing that refactor in another PR.

PR #541, opened 2018-10-10
refactor: travis build

### LINKED ISSUE #517: refactor CI config
### Problem
Right now the, the travis.yml config breaks on make ipfs with write errors. This is a strange, runtime anomaly that occurs when Docker fails to write to the storage driver . We do not have control over this as the build environment is managed by Travis.

We also have  some things added into the travis environment, which may not be needed, since these are handled in the Docker image itself:
```
before_install:
  - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
  - sudo apt-get update
  - sudo apt-get install gcc-5 g++-5
  - sudo rm /usr/bin/g++
  - sudo ln -s /usr/bin/g++-5 /usr/bin/g++
```

### Solution
Refactor the entire travis config, and leverage docker-compose or a more a sane approach to running the backing services than with make recipes.

seeing as we only perform a lint and test coverage in the script, we can simplify the build config drastically.

```
script:
  - npm run lint
  - npm run coverage
```

And make it more future proof for when we have integration/unit tests in place. Rightnow from what I can, node only really runs the `dist/babel/test/PrepareDB.js` which prepares mongo and one function `async function insertTestWorks` that writes to the db. 

I think we should re-order the travis config, to better handle the starting up of services as containers, the test DB preparation, the lint and coverage, and resolve the weird as hell make errors we get sporadically. 

---
Moved from poetapp/random#177
/cc @WesleyCharlesBlake

warrenv:

### Release

## Description of Changes

Closes #517 

This commit:
  - Refactors the travis CI build to use docker containers.
  - Prepares for running functional tests.


.travis.yml
```diff
@@ -7,36 +8,38 @@ notifications:
...
-  - npm run test-prepare-db
-  - sleep 5
-  - npm start &
-  - sleep 5
+
+install:
+  - DC_FILES="-f docker-compose.yml -f docker-compose.ci.yml"
+  - docker-compose $DC_FILES build
+  - docker images
+  - docker-compose $DC_FILES up -d
```
lautarodragan: I would have thought this would stop the logs from showing up in the Travis report, but looks like not... any idea why?

warrenv:
@WesleyCharlesBlake I would like to fine-tune the docker builds in a separate PR. The scope of this small PR is to get the current tests running via containers. After that, the plan is to add the functional tests. After that, make it more efficient. This PR is part of several small steps in that direction. Do you agree with that approach?

warrenv:
@lautarodragan Eliminating the makefile will be a separate PR.

PR #548, opened 2018-10-11
HealthEndpoint: Bitcoin Core Information

jwicks31:

### Release

## Health Module
Added a Health Module with the following responsibilities:

### Health
Entry Point for the Module. Initializes the container and starts the HealthService.

### Health Service
At a set interval checks IPFS Connection and gathers Bitcoin-Core information. Storing each in the database.

### HealthController
Supplies HealthModule with methods to check IPFS and gather Bitcoin-Core information. 

## API Module
Changes to the API Module:

### HealthController
Added the following information provided by querying the Health Mongo Database
- bitcoin core getblockchaininfo 
- - blocks
- - verificationprogress
- - bestblockhash
- - warnings
- - size_on_disk
- bitcoin core getnetworkinfo 
- - version
- - subversion
- - connections
- - networkactive
- - protocolversion
- - warnings
- bitcoin core getwalletinfo 
- - balance
- - txcount

Part of #376


src/API/HealthController.ts
```diff
@@ -30,22 +52,60 @@ export class HealthController {
     }
   }
 
-  private async checkIPFS(): Promise<boolean> {
+  private async checkIPFS(): Promise<any> {
     try {
-      const ipfsConnection = await this.ipfs.getVersion()
-      return isStatus200(ipfsConnection)
+      const { ipfsIsConnected } = (await this.collection.findOne({ name: 'ipfsConnected' })) || {
```
kennylavender: Please put all mongo logic in a DAO, see StorageWriter. Same goes for all `this.collection` calls.

jwicks31:
@kennylavender I updated the branch to include only the new logic and not the refactor. Is this better in terms of PR size?

src/API/HealthController.ts
jwicks31: Ok. Is this something I should do in another PR then?

kennylavender: Yup, I was just pointing it out so you are aware for next time 👍

kennylavender:
@jwicks31 much better, now we can focus on reviewing just the new logic instead of the new logic and the refactor 👍 

I might have refactored first, then added the new logic. But I am fine with this as long as you follow up with a refactor PR.

tests/functional/health.ts
```diff
@@ -27,20 +27,41 @@ describe('Health Endpoint Returns the Correct Fields', async (assert: any) => {
...
       expected: 'boolean',
     })
 
     assert({
-      given: 'a request to the health endpoing',
+      given: 'a request to the health endpoint',
       should: 'return object with property ipfsIsConnected',
-      actual: typeof ipfsIsConnected,
-      expected: 'boolean',
+      actual: ipfsIsConnected !== null,
```
kennylavender: use ramda isNill for these checks so that it also checks for undefined.

https://ramdajs.com/docs/#isNil

src/Health/HealthServiceConfiguration.ts
```diff
@@ -0,0 +1,3 @@
+export interface HealthServiceConfiguration {
```
kennylavender: Move interface into the class file.

src/Health/HealthService.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.logger.info('Health Cron Starting...')
+    this.interval.start()
+  }
+
+  async stop() {
+    this.logger.info('Health Cron Stopping...')
+    this.interval.stop()
+  }
+
+  private getHealth = async (): Promise<any> => {
```
kennylavender: Promise<void> as it does not return anything.

src/Health/HealthConfiguration.ts
```diff
@@ -0,0 +1,8 @@
+import { LoggingConfiguration, BitcoinRPCConfiguration } from 'Configuration'
```
kennylavender: Move everything in this file to the Health.ts file as it for that class.

src/Health/HealthService.ts
kennylavender: Also, see the StorageWriter. This method should fire a rabbitmq message instead of calling the controller directly.

src/Health/HealthController.ts
```diff
@@ -0,0 +1,106 @@
...
+      {
+        $set: {
+          walletInfo,
+        },
+      },
+      { upsert: true }
+    )
+  }
+
+  async getNetworkInfo(): Promise<void> {
```
kennylavender: refreshNetworkInfo might be a better name.

Also split this method up into multiple private methods. remember each function should do as close to 1 thing as possible.

There can be a private method that only calls and returns the result of `this.bitcoinCore.getNetworkInfo()`. I might call this method getNetworkInfo, hence the better name of this containg method.

src/Health/HealthController.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    const {
+      version,
+      subversion,
+      connections,
+      networkactive,
+      protocolversion,
+      warnings,
+    } = await this.bitcoinCore.getNetworkInfo()
+    const networkInfo = { version, subversion, connections, networkactive, protocolversion, warnings }
```
kennylavender: How come your destructuring all these properties, then reassigning them to a new object in exactly the same structure?

src/Health/HealthController.ts
kennylavender: See the StorageWriter, see if you can make updateNetworkInfo a pipe method something like: 

```ts
refreshNetworkInfo = pipe(
  this.log(LogTypes.trace)('refreshing network info'),
  this.getNetworkInfo,
  this.log(LogTypes.trace)('new info gathered, saving network info'),
  this.updateNetworkInfo,
  this.log(LogTypes.trace)('refreshed network info')
)

File: src/Health/HealthController.ts [hunk 8]
```diff
@@ -0,0 +1,106 @@
...
+      {
+        $set: {
+          blockchainInfo,
+        },
+      },
+      { upsert: true }
+    )
+  }
+
+  async getWalletInfo(): Promise<void> {
```
kennylavender: same adjustments as getNetworkInfo.

src/Health/HealthController.ts
```diff
@@ -0,0 +1,106 @@
...
+    @inject('DB') db: Db,
+    @inject('BitcoinCore') bitcoinCore: BitcoinCore
+  ) {
+    this.logger = childWithFileName(logger, __filename)
+    this.db = db
+    this.collection = this.db.collection('health')
+    this.bitcoinCore = bitcoinCore
+  }
+
+  async getBlockchainInfo(): Promise<void> {
```
kennylavender: same adjustments as getNetworkInfo.

src/API/HealthController.ts
```diff
@@ -39,13 +66,49 @@ export class HealthController {
...
   }
 
+  private async getBlockchainInfo(): Promise<any> {
+    try {
+      const { blockchainInfo } = (await this.collection.findOne({ name: 'blockchainInfo' })) || {
+        blockchainInfo: 'Getting Blockchain Info...',
+      }
+      return blockchainInfo
+    } catch (e) {
+      return e
```
kennylavender: I am not sure we will want to expose the error through the api like this. @lautarodragan thoughts on what you would like to have returned?

jwicks31:
@kennylavender ready for another review whenever you get a chance!

src/API/HealthController.ts
jwicks31: I changed these to return 'Error retrieving `x`'

src/API/HealthController.ts
```diff
@@ -39,13 +66,49 @@ export class HealthController {
     }
   }
 
+  private async getBlockchainInfo(): Promise<any> {
+    try {
+      const { blockchainInfo } = (await this.collection.findOne({ name: 'blockchainInfo' })) || {
```
kennylavender: I think we can just default it to an empty object, don't think we need a message.

```ts
const { blockchainInfo = {} } = await this.collection.findOne({ name: 'blockchainInfo' })
```

src/API/HealthController.ts
```diff
@@ -39,13 +66,49 @@ export class HealthController {
...
+        blockchainInfo: 'Getting Blockchain Info...',
+      }
+      return blockchainInfo
+    } catch (e) {
+      return { blockchainInfo: 'Error retrieving blockchainInfo...' }
```
kennylavender: I think returning blockchainInfo inside an object is going to result in a nested blockchainInfo

```ts
{
  blockchainInfo: {
    blockchainInfo: 'Error retrieving blockchainInfo...'
  }
}
```

We can just change this key to error like:

```ts
return { error: 'Error retrieving blockchainInfo...' }
```

So that it puts an error prop instead

```ts
{
  blockchainInfo: {
    error: 'Error retrieving blockchainInfo...'
  }
}
```

kennylavender: If you agree, apply to the other 2 functions as well

src/API/HealthController.ts
kennylavender: If you agree, apply to the other 2 functions as well

src/API/HealthController.ts
jwicks31: Ok yeah. I was thinking I was destructuring blockchainInfo from the returned value of this function, but I'm not.

src/API/HealthController.ts
```diff
@@ -39,13 +66,43 @@ export class HealthController {
     }
   }
 
+  private async getBlockchainInfo(): Promise<any> {
```
kennylavender: These can now be changed to return `Promise<object>`

src/Health/HealthService.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.logger.info('Health Cron Starting...')
+    this.interval.start()
+  }
+
+  async stop() {
+    this.logger.info('Health Cron Stopping...')
+    this.interval.stop()
+  }
+
+  private getHealth = async (): Promise<void> => {
```
kennylavender: This is still calling the controller directly when it should be publishing a rabbit mq message that the router listens for, but I am ok with doing that in a refactor PR if you would rather.

jwicks31: Ok. I'll do this in the refactor PR.

Issue #435, opened 2018-09-13
Add Wallet Balance Alert to Health Endpoint


lautarodragan:
- [ ] Add wallet balance alert threshold to the detailed health endpoint (@jwicks) when the balance drops below a pre-defined (env var) bitcoin amount.

lautarodragan:
Forgot to add a description to this one at the moment I created it...

I believe this one came out of the bitcoin-core meeting, and it was about the node throwing errors when there are insufficient funds in the wallet, rather than correctly detecting this and logging it.

geoffturk:
Yup, that was the context.

jwicks31:
Moved the last part of this issue to [a new](https://github.com/poetapp/node/issues/629) issue

PR #566, opened 2018-10-16
Insufficient Funds Error

### LINKED ISSUE #565: Detect and Log Create Claim Insufficient Funds Error
- Detect and log in Po.et node logs when the node can no longer post claims because the bitcoind wallet does not have sufficient funds to write the batch hash to the Bitcoin mainnet/testnet blockchains.

jwicks31:

### Release

## Description of Changes

Resolves #565 Par of #435 

- Moved `bitcoinCore.fundRawTransaction` to fundRawTransactions method on the BlockchainWriter controller and added a try/catch around `bitcoinCore.fundRawTransaction`
- Added `InsufficientFundsException` to `Exceptions`


src/BlockchainWriter/Controller.ts
```diff
@@ -90,8 +91,19 @@ export class Controller {
     })
   }
 
-  private anchorData = async (data: string) => {
+  private fundTransaction = async (rawTransaction: string) => {
```
lautarodragan: This can be simplified by using [`rethrow`](https://github.com/poetapp/node/blob/master/src/Extensions/Promise.ts#L6):

```suggestion
// outside the class, as a pure function
const translateFundTransactionError = async (error: Error) => 
  exception.message === 'Insufficient funds'
    ? new InsufficientFundsException('Insuffiecient funds for transaction')
    : error
```

```ts
// in the class
private anchorData = async (data: string) => {
  /// ...
  const fundedTransaction = await bitcoinCore.fundTransaction(rawTransaction).rethrow(translateFundTransactionError)
```

src/BlockchainWriter/Controller.ts
```diff
@@ -90,8 +91,19 @@ export class Controller {
...
     const { bitcoinCore } = this
+    try {
+      return await bitcoinCore.fundRawTransaction(rawTransaction)
+    } catch (exception) {
+      if (exception.message === 'Insufficient funds')
```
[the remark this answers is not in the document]
lautarodragan: @kennylavender see my comment — the pure function could be unit tested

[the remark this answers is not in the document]
kennylavender: @lautarodragan Cool, that works for me

src/BlockchainWriter/Controller.ts
jwicks31: I'm getting this error when trying to use rethrow: 
```
module: "BlockchainWriter"
    file: "Controller"
    method: "anchorNextHash"
    ipfsDirectoryHash: "QmWy3kwndywtopWB2audW12HFnAe2gcQe3MupdSV8dNKX1"
    exception: {
      "stack": [
        "TypeError: bitcoinCore.fundRawTransaction(...).rethrow is not a function",
        "    at Controller.anchorData (/usr/src/app/dist/babel/src/BlockchainWriter/Controller.js:36:92)",
      ],
      "message": "bitcoinCore.fundRawTransaction(...).rethrow is not a function",
      "type": "TypeError"
    }
```

kennylavender: @jwicks31 I think you might need to import the rethrow promise extensions. src/Extensions/Promise.ts

jwicks31: It looks like they are imported at the app level. Should that cover it?

kennylavender: Ah, yeah if its already imported then you shouldn't need to import again. Can you find another example of rethrow somewhere else being used in the codebase?

jwicks31: `src/StorageReader/ClaimController.ts` uses it.

kennylavender: Interesting, maybe bitcoinCore.fundRawTransaction does not return an es6 Promise, but a promise from a custom library or something, that is the only thing I can think of that the global es6 Promise is not being extended.

kennylavender: Maybe wrap the call with another promise so you can use rethrow or wrap it with another function thats only responsibility is the try/catch for now. I don't have time to investigate why rethrow is not working on bitcoinCore.fundRawTransaction

lautarodragan: Darn, it uses bluebird. 

We can refactor `rethrow` in the future so it takes a Promise in a curried way and can be used in pipes.

For now I think this is best: 

```ts
const translateFundTransactionError = (error: Error) => {
  throw error.message === 'Insufficient funds'
    ? new InsufficientFundsException('Insuffiecient funds for transaction')
    : error
}

const fundedTransaction = await bitcoinCore.fundRawTransaction(rawTransaction).catch(translateFundTransactionError)

```

PR #570, opened 2018-10-16
Refactor Health Service/Health Router

### LINKED ISSUE #569: Move Health Controller to Router and use RabbitMQ
Moved the calls to the HealthController to a router and add RabbitMQ messaging to health service/router.

jwicks31:

### Release

## Description of Changes

Resolves #569 

Refactors Health Service to publish messages, which are consumed by the router. The router then calls the HealthController methods to refresh the information.


src/Health/HealthController.ts
```diff
@@ -86,7 +86,7 @@ export class HealthController {
     return blockchainInfo
   }
 
-  async getWalletInfo(): Promise<WalletInfo> {
+  private async getWalletInfo(): Promise<WalletInfo> {
```
kennylavender: I was trying to review this PR from my phone earlier, but this change confused me because it's not related to the changes here so I had to wait until I could get to my laptop.

You could fix things like this in a separate PR next time, just a tip to make things simpler for reviewers :)

src/Health/Router.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.messaging = messaging
+    this.controller = controller
+    this.exchange = exchange
+  }
+
+  async start() {
+    await this.messaging.consume(this.exchange.getHealth, this.onGetHealth)
+  }
+
+  async stop() {
```
lautarodragan: It's preferred for the root of the module to do the `this.messaging.stop()`, since it's the one doing the `messaging.start()`.

The reason our RMQ routers have a `start` but no `stop` is because [we're not cancelling channels](https://www.squaremobius.net/amqp.node/channel_api.html#channel_cancel).

jwicks31: Ok. Should I keep?:
```js
async stop() {
    this.logger.info('Stopping Health Messaging...')
  }
```

PR #573, opened 2018-10-16
OP_RETURN Fix

### LINKED ISSUE #551: Error when reading OP_RETURN data unrelated to POET
https://github.com/poetapp/node/blob/8e19bce6d2cbecebe3b2ab465c46f4d933d59660/src/BlockchainReader/Bitcoin.ts#L45-L57

When reading data from the blockchain, nothing's validating that the length of the data stored in the OP_RETURN output.

This leads to `dataToPoetAnchor` trying to read bytes outside the range of the buffer, which throws:

```
    method: "scanNewBlock"
    error: {
      "stack": [
        "RangeError [ERR_OUT_OF_RANGE]: The value of \"offset\" is out of range. It must be >= 0 and <= 4. Received 6",
        "    at boundsError (internal/buffer.js:53:9)",
        "    at Buffer.readInt8 (internal/buffer.js:309:5)",
        "    at Object.exports.dataToPoetAnchor.data [as dataToPoetAnchor] (/usr/src/app/dist/babel/src/BlockchainReader/Bitcoin.js:27:36)",
        "    at dataOutputToPoetTransactionAnchor (/usr/src/app/dist/babel/src/BlockchainReader/Bitcoin.js:16:28)",
        "    at transactionToPoetAnchor (/usr/src/app/dist/babel/src/BlockchainReader/Bitcoin.js:11:26)",
        "    at Array.map (<anonymous>)",
        "    at Object.exports.blockToPoetAnchors.block [as blockToPoetAnchors] (/usr/src/app/dist/babel/src/BlockchainReader/Bitcoin.js:7:48)",
        "    at ClaimController.scanBlock (/usr/src/app/dist/babel/src/BlockchainReader/ClaimController.js:46:35)",
      ],
      "message": "The value of \"offset\" is out of range. It must be >= 0 and <= 4. Received 6",
      "code": "ERR_OUT_OF_RANGE",
      "type": "NodeError"
    }
    blockHeight: 1280418
    lastBlockHeight: 1280417
```

We could solve this by having `dataToPoetAnchor` return a `undefined`. 

```ts
export const dataToPoetAnchor = (data: string): PoetAnchor => {
  const buffer = Buffer.from(data, 'hex')
  if (buffer.byteLength < 4 + 2 + 1 + N)
    return undefined
  const prefix = buffer.slice(0, 4).toString()
  const version = Array.from(buffer.slice(4, 6))
  const storageProtocol = buffer.readInt8(6)
  const ipfsDirectoryHash = buffer.slice(7).toString()
  return {
    storageProtocol,
    prefix,
    version,
    ipfsDirectoryHash,
  }
}
```

Where `N` is a dynamic value that depends on the storageProtocol and multihash. We're only supporting one version of multihash and storageProtocol at the moment, so it's safe to hard-code it, but we'll want to use a constant rather than a magic number.

__Note__

Returning `Option<PoetAnchor>` would be nicer but we don't have anything to support Option in the code base yet, and it'd require a bit of refactor.

Returning undefined will solve the issue immediately since `blockToPoetAnchors` has a `.filter(isTruthy)`.

__FYI__

The block is [1280418](https://test-insight.bitpay.com/api/block/00000000000012ef158b07cb514f14156cf99a628eff53f789db05af1a1bbd88).

It has three transactions with data:
- [81120518637a0b964da9f9075995335eb12397100e5f7f9a6e711c16b5f292f6](https://test-insight.bitpay.com/api/tx/81120518637a0b964da9f9075995335eb12397100e5f7f9a6e711c16b5f292f6)
- [918cc48177c39a01ab6a20d3e980526775262996de00928fa92d163bd2508812](https://test-insight.bitpay.com/api/tx/918cc48177c39a01ab6a20d3e980526775262996de00928fa92d163bd2508812)
- [7fa11a4809441799f6392fe06c1ba13dca53e72b5524f5be23d4560262cc9b67](https://test-insight.bitpay.com/api/tx/7fa11a4809441799f6392fe06c1ba13dca53e72b5524f5be23d4560262cc9b67)

None of those is a Po.et anchor.

`918cc48177c39a01ab6a20d3e980526775262996de00928fa92d163bd2508812` is probably the one throwing the stack trace shown above:

```js
> Buffer.from('1437681861', 'hex')
<Buffer 14 37 68 18 61>
```

jwicks31:

### Release

## Description of Changes

Fixes #551


src/BlockchainReader/Bitcoin.test.ts
```diff
@@ -184,10 +184,17 @@ describe('Bitcoin.dataToPoetAnchor', async assert => {
...
   }
 
-  assertSample('POET', [2, 3], StorageProtocol.IPFS, 'Jim')
-  assertSample('BARD', [2, 3], StorageProtocol.IPFS, 'Robert')
-  assertSample('POET', [1, 0], StorageProtocol.IPFS, 'Roger')
-  assertSample('BARD', [1, 0], StorageProtocol.IPFS, 'Syd')
+  assert({
+    given: 'A non Poet OP_RETURN Data',
+    should: 'return undefined',
+    actual: dataToPoetAnchor('918cc48177c39a01ab6a20d3e980526775262996de00928fa92d163bd2508812'),
```
warrenv: When reading this test, it's not obvious to me why this data is incorrect.
Can the 'given' be expanded on or the argument made to be easier to see that it's incorrect?

src/BlockchainReader/Bitcoin.ts
```diff
@@ -44,6 +56,7 @@ const dataOutputToData = (dataOutput: VOutWithTxId): string => {
 
 export const dataToPoetAnchor = (data: string): PoetAnchor => {
   const buffer = Buffer.from(data, 'hex')
+  if (buffer.byteLength < anchorByteLength) return undefined
```
kennylavender: don't really care for the idea of this function returning `undefined || PoetAnchor`.

I think some refactoring, then making the change is in order to simplify these functions, to allow us to add this check easier.

take a look at blockToPoetAnchors

```ts
export const blockToPoetAnchors = (block: Block): ReadonlyArray<PoetBlockAnchor> =>
  block.tx
    .map(transactionToPoetAnchor)
    .filter(isTruthy)
    .filter(poetAnchorHasCorrectPrefix)
    .map(poetAnchorWithBlockData(block))
```

We can refactor and fix some of the functions it uses to be simpler by making them responsible for a single thing

Something like this

```ts
export const blockToPoetAnchors = (block: Block): ReadonlyArray<PoetBlockAnchor> =>
  block.tx
    .map(transactionToDataOutput)
    .filter(isTruthy)
    .map(dataOutputToData)
    .map(dataToBuffer)
    .map(bufferToPoetAnchor)
    .filter(poetAnchorHasCorrectPrefix)
    .map(poetAnchorWithBlockData(block))
```

That allows us to then add a filter function that checks the buffer length.

```ts
export const blockToPoetAnchors = (block: Block): ReadonlyArray<PoetBlockAnchor> =>
  block.tx
    .map(transactionToDataOutput)
    .filter(isTruthy)
    .map(dataOutputToData)
    .map(dataToBuffer)
    .filter(isCorrectBufferLength) // we can now simply add this new filter
    .map(bufferToPoetAnchor)
    .filter(poetAnchorHasCorrectPrefix)
    .map(poetAnchorWithBlockData(block))
```

Refactoring in a separate PR first, then make the change.

src/BlockchainReader/Bitcoin.ts
```diff
@@ -8,6 +8,18 @@ interface VOutWithTxId extends VOut {
   readonly transactionId: string
 }
 
+enum AnchorSectionsLengths {
+  Prefix = 4,
+  StorageProtocol = 1,
+  Version = 2,
+  IPFSHash = 46,
+}
+const anchorByteLength =
```
kennylavender: instead of adding these together manually can we sum the values of the enum?

lautarodragan: maybe Object.values

src/BlockchainReader/Bitcoin.ts
jwicks31: Ok. Would you be OK with approving this PR and I will open one with the refactor? This is blocking the QA environment testing I believe.

src/BlockchainReader/Bitcoin.ts
jwicks31: We could do either(or something like it):
```ts
const AnchorSectionsLengths = {
  Prefix: 4,
  StorageProtocol:  1,
  Version: 2,
  IPFSHash: 46,
}

const add = (a: number, b: number) => a + b
const anchorBytesLength = Object.values(AnchorSectionsLengths).reduce(add)
```
or:
```ts
enum AnchorSectionsLengths {
  Prefix = 4,
  StorageProtocol =  1,
  Version = 2,
  IPFSHash = 46,
}

const addNums = (a: number, b: number) => isNaN(a) || isNaN(b) ? a + 0 : a + b
const anchorByteLength = Object.values(AnchorSectionsLengthss).reduce(addNums, 0)
```

src/BlockchainReader/Bitcoin.ts
kennylavender: I am going to say no unless other disagree. I would like to do the refactor first, then the fix. We have a bad habit of rushing changes to code because its blocking something else, in reality, something is almost always blocking something else, we need to start tackling issues better at some point and I think we should start now so we start forming those better habits.

src/BlockchainReader/Bitcoin.ts
kennylavender: Better, it might be a bit nicer using ramda though

https://ramdajs.com/docs/#values
https://ramdajs.com/docs/#sum

```ts
const sumObjectValues = pipe(values, sum)
const anchorByteLength = sumObjectValues(AnchorSectionsLengths)
```

PR #595, opened 2018-10-19
chore(scripts): clean output of npm run create-claim

lautarodragan:

### Release

## Description of Changes

Removes some of the output so the entire output of the script is JSON when run with `--silent`, which allows us to manage the output with `jq` to do things like:

```
$ WORK_ID=$(npm run --silent create-claim | jq .id)
```


warrenv:
NOTE: jq is not installed in the container so the command referenced in the description will not work.

PR #602, opened 2018-10-19
chore: remove enableTimestamping option

lautarodragan:

### Release

## Description of Changes

I originally included this option because the node needed to be configured with an address and a private key and would throw errors otherwise.

Now that it is using the Bitcoin Core wallet this is no longer the case, and this option has caused some confusion in the past.


warrenv:
I would like to keep this option.

It is currently being used in the integration tests - they rely on the default value of false because they currently don't have a need to write to the bitcoind-1 service. 

I can also see a use case where someone wants to run a "read-only" version of node. I guess that could be done by not putting any funds in the wallet, but it would generate a bunch of errors.

lautarodragan:
@warrenv good catch, I had not thought of the tests that create claims but can't yet anchor them.

Issue #490, opened 2018-09-27
Anchor optimization: multi hash as string


lautarodragan:
We're currently storing the multihash returned by IPFS as a plain string.

```ts
export const poetAnchorToData = (poetAnchor: PoetAnchor) => {
  if (poetAnchor.prefix.length !== 4) throw new Error('Prefix should be a string of length 4')
  if (poetAnchor.version.length !== 2) throw new Error('Version should be an array of length 2')
  return Buffer.concat([
    Buffer.from(poetAnchor.prefix),
    Buffer.from([...poetAnchor.version]),
    Buffer.from([poetAnchor.storageProtocol]),
    Buffer.from(poetAnchor.ipfsDirectoryHash), // <==== here
  ]).toString('hex')
}
```

This multihash is actually a base58 representation of a byte buffer. Treating it as an utf8 actually wastes a couple of bytes.

```js
const bs58 = require('bs58')

const ipfsDirectoryHash = 'QmaKtaPqfWvss2ndhGdcLRc9uGcLvUUyuZeGcrDZWN58Zi'

const multihashBuffer = bs58.decode(ipfsDirectoryHash)
const stringBuffer = Buffer.from('QmaKtaPqfWvss2ndhGdcLRc9uGcLvUUyuZeGcrDZWN58Zi')

console.log('base58', multihashBuffer.byteLength, multihashBuffer)
console.log('string', stringBuffer.byteLength, stringBuffer)

// outputs
// base58 34 <Buffer 12 20 b2 1a 57 80 ff 87 20 4a a3 f8 ae 88 5e 93 52 dc 3e d8 a5 72 8a 56 8e bd 82 a0 fd 4b 47 3e 3a 75>
// string 46 <Buffer 51 6d 61 4b 74 61 50 71 66 57 76 73 73 32 6e 64 68 47 64 63 4c 52 63 39 75 47 63 4c 76 55 55 79 75 5a 65 47 63 72 44 5a 57 4e 35 38 5a 69>
```
Replacing `Buffer.from(poetAnchor.ipfsDirectoryHash)` with `bs58.decode(ipfsDirectoryHash)` should do the trick.

```ts
export const poetAnchorToData = (poetAnchor: PoetAnchor) => {
  if (poetAnchor.prefix.length !== 4) throw new Error('Prefix should be a string of length 4')
  if (poetAnchor.version.length !== 2) throw new Error('Version should be an array of length 2')
  return Buffer.concat([
[...]

PR #609, opened 2018-10-22
BREAKING CHANGE: Anchor optimization, ipfsDirectoryHash

wzalazar:

### Release

## Description of Changes

Fix https://github.com/poetapp/node/issues/490

This PR introduces:

- Handling the `ipfsDirectoryHash` like as base58

like as you can see I have to change the `IPFSHash` on `AnchorSectionsLengths` to:

```
 - IPFSHash: 46,
 + IPFSHash: 34,
```

because:

```
const multihashBuffer = bs58.decode(ipfsDirectoryHash)
const stringBuffer = Buffer.from('QmaKtaPqfWvss2ndhGdcLRc9uGcLvUUyuZeGcrDZWN58Zi')

console.log('base58', multihashBuffer.byteLength, multihashBuffer)
console.log('string', stringBuffer.byteLength, stringBuffer)

// outputs
// base58 34 <Buffer 12 20 b2 1a 57 80 ff 87 20 4a a3 f8 ae 88 5e 93 52 dc 3e d8 a5 72 8a 56 8e bd 82 a0 fd 4b 47 3e 3a 75>
// string 46 <Buffer 51 6d 61 4b 74 61 50 71 66 57 76 73 73 32 6e 64 68 47 64 63 4c 52 63 39 75 47 63 4c 76 55 55 79 75 5a 65 47 63 72 44 5a 57 4e 35 38 5a 69>
```

`Pay attention please is a BREAKING CHANGE`


src/BlockchainReader/Bitcoin.test.ts
```diff
@@ -25,7 +25,7 @@ describe('Bitcoin.blockToPoetAnchors', async assert => {
     expected: true,
   })
 
-  const ipfsDirectoryHashes = ['QmaKtaPqfWvss2ndhGdcLRc9uGcLvUUyuZeGcrDZWN58Zi']
+  const ipfsDirectoryHashes = ['9tmyge2RGfyymosruzDgPjdHNXpFmp5BqY5fXNqskw4PitFPuZaKQVcdcodDqPE']
```
wzalazar: @lautarodragan I'm trying to figure out why I have to change the value here.

lautarodragan: I think it's because this is a breaking change and the tests are grabbing test data that is no longer valid.

```ts
import * as TestBlock from './TestData/block-0000000070746b06bbec07a7cd35e0c6d47bfa4e2544a6a1d2aa6efc49d47679.json'

const poetAnchors = blockToPoetAnchors(TestBlock as any) // as any: see footer note
const anchorsIpfsHashes = poetAnchors.map(poetAnchor => poetAnchor.ipfsDirectoryHash)
const ipfsDirectoryHashes = ['9tmyge2RGfyymosruzDgPjdHNXpFmp5BqY5fXNqskw4PitFPuZaKQVcdcodDqPE']
const anchorHasUnexpectedIpfsHash = (poetAnchor: PoetBlockAnchor) =>
  !ipfsDirectoryHashes.includes(poetAnchor.ipfsDirectoryHash)
```

Then these two tests compare the new code agains the old test data, which fails:

```ts
    assert({
      given,
      should: 'all returned elements should have an expected IPFS Directory Hash',
      actual: poetAnchors.find(anchorHasUnexpectedIpfsHash),
      expected: undefined,
    })
// ...
    assert({
      given,
      should: 'all expected IPFS Directory hashes should be in the returned anchors',
      actual: equals(anchorsIpfsHashes.sort(localeCompare), ipfsDirectoryHashes.sort(localeCompare)),
      expected: true,
    })
```

So, TDD-wise, you shouldn't be changing this line of code and the tests should fail, initially.

Then, to fix the failing tests, you'll need new test data.

I grabbed this test data originally by running new code on testnet and downloading the block in which my test transactions got confirmed. 

The problem with this approach is that we could be introducing a bug and kind of "forcing" the tests to pass, so it requires paying close attention to the changes (to both the PR creator and the reviewers). I'm open to better alternatives, but this change is pretty small so downloading a new testnet block can be enough.

src/BlockchainReader/Bitcoin.ts
```diff
@@ -12,7 +13,7 @@ const AnchorSectionsLengths = {
   Prefix: 4,
   StorageProtocol: 1,
   Version: 2,
-  IPFSHash: 46,
+  IPFSHash: 34,
```
lautarodragan: Can you add something like this to the PR description to help reviewers?

```js
console.log('before', Buffer.byteLength(bs58.decode('QmaKtaPqfWvss2ndhGdcLRc9uGcLvUUyuZeGcrDZWN58Zi')))
console.log('after', Buffer.byteLength(Buffer.from('QmaKtaPqfWvss2ndhGdcLRc9uGcLvUUyuZeGcrDZWN58Zi')))
```

typings/bs58.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'bs58'
```
lautarodragan: Can you add type declarations for `encode` and `decode`?

src/BlockchainWriter/Bitcoin.test.ts
```diff
@@ -46,21 +49,26 @@ describe('Bitcoin.getData', async assert => {
       prefix: PREFIX_POET,
       version: [0, 2],
       storageProtocol: StorageProtocol.IPFS,
-      ipfsDirectoryHash: 'ipfsDirectoryHash',
+      ipfsDirectoryHash: getB58('ipfsDirectoryHash'),
     }
 
     testGetData(poetAnchor)
-    testGetData({ ...poetAnchor, prefix: PREFIX_BARD, ipfsDirectoryHash: 'another ipfsDirectoryHash' })
+    testGetData({ ...poetAnchor, prefix: PREFIX_BARD, ipfsDirectoryHash: getB58('another ipfsDirectoryHash') })
```
lautarodragan: I don't think you really need to run `getB58` on this input...

If it's failing because it tries to parse it as a BS58 string, can you just change it to an actual IPFS hash instead?

typings/bs58.d.ts
wzalazar: yes, sure. I tried to use the official typings but it does not comply with the current specifications ;)

src/BlockchainReader/Bitcoin.ts
wzalazar: yes, sure

src/BlockchainReader/Bitcoin.test.ts
wzalazar: yes, this is my concern. I'm trying to understand all around this test

typings/bs58.d.ts
wzalazar: done!

src/BlockchainWriter/Bitcoin.test.ts
wzalazar: Yes, it fails because tries to read BS58 string.

`UnhandledPromiseRejectionWarning: Error: Non-base58 character`

ok, I will change for a real IPFS hash.

wzalazar:
@jwicks31 awesome code review! I will expect another code review before to merge it. Can you check @lautarodragan when you have a chance?

PR #612, opened 2018-10-22
feature: implement POST /files endpoint

### LINKED ISSUE #611: Add POST /file endpoint
related #260 

will add more details here soon

kennylavender:

### Release

## Description of Changes

closes #611 

This pr ended up being larger than needed because we don't have any kind of feature toggles in place #685.

Add the ability for users to post files to IPFS through the node.

- [x] store the resulting hash in the database
- [x] handle multiple file uploads
- [x] use multipart
- [x] use streams
- [x] returns an array of objects that contain a hash
- [x] use configuration for ipfs url
- [x] update readme

## Out of Scope / Todo after this PR
- throw error if no files passed in via multipart
- recheck to see if we can pass through the ReadStream without saving a temp file.  Tried this originally in #682 , but would like to double check to see if its possible.
- adjust functional tests to use the file endpoint
- publish messages that contain the hashes of the uploaded files
- handle larger file uploaded with other uploaded techniques than multipart ( thx @krobi64 )
  - IPFS currently only takes multipart if I remember correctly, but they were working on a better way for larger files.
  - https://gist.github.com/CMCDragonkai/6bfade6431e9ffb7fe88#a-note-about-nodejs
  - https://stackoverflow.com/questions/25810051/filereader-api-on-big-files/25813769#25813769

## Manual Test Directions
* terminal 1 ```cp docker-compose.override.yml.exampl docker-compose.override.yml```
* edit docker-compose.override.yml and uncomment poet-node's `volumes` and `command`, save.
* terminal 1 ```docker-compose up -d```
* terminal 1 ```docker-compose exec poet-node bash```
* terminal 1 ```docker-compose npm i && npm run build && npm run start```
* terminal 2 ```docker-compose exec mongo bash```
* terminal 2 ```mongo --username poet --password poetPass poet```
* terminal 3 ```echo 'my file' > text.txt```
* terminal 3 ```curl -F file=@text.txt "http://localhost:18080/files"```   
   should result in something like:  [{"hash":"QmRDfwqbJocLAEdxeVjLbUeXPKpNcj3n9GrPUvukbvfaF5"}]
* terminal 2 ```db.files.find()```   
  should result in something like: { "_id" : ObjectId("5bda22c59cb490008e6abb86"), "hash" : "QmRDfwqbJocLAEdxeVjLbUeXPKpNcj3n9GrPUvukbvfaF5" }


lautarodragan:
@kennylavender @warrenv if it's not too complicated and is still valid for testing, I'd rather have a script that streams randomly generated data than adding 4MB of test data to the repo.

lautarodragan:
That works for now!

tests/fixtures/files/text-utf8-non-ascii.txt
```diff
@@ -0,0 +1 @@
+Po.et está creando un libro de cuentas abierto y universal que registra información inmutable y con marca de tiempo sobre su contenido creativo y utiliza nuestros protocolos abiertos diseñados para la interoperabilidad con los estándares actuales de la industria en medios y publicaciones.
```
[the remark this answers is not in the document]
kennylavender: Thanks, I think il add more languages to this file.

tests/integration/API/post-file.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  // Setup Mongodb and the app server
+  const { db, server } = await setUpServerAndDb({ PREFIX, NODE_PORT })
+
+  {
+    const given = 'a utf8 ascii text file and post request'
+    const should = 'return a hash that can be used to verify the file'
+    const file = fs.readFileSync(files.utf8AsciiText).toString()
+    const fileStream = fs.createReadStream(files.utf8AsciiText)
+
+    try {
```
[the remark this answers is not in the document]
kennylavender: I like that idea, that will work if we only do 1 assert per set of network requests. I wasn't sure how I want to handle the other assertions, http status for example. If we use pipe to build up the test functions to wrap try around then we have to make the same network calls per assertion, but it would look much nicer. I think il try it out.

src/API/Router.ts
```diff
@@ -74,7 +78,7 @@ export class Router {
     this.koa.use(KoaCors({ expose: ['X-Total-Count'] }))
     this.koa.use(LoggerMiddleware(this.logger))
     this.koa.use(HttpExceptionsMiddleware)
-    this.koa.use(KoaBody({ textLimit: 1000000 }))
+    this.koa.use(KoaBody({ textLimit: 1000000, multipart: true }))
```
krobi64: Don't think this is needed now, but we should consider some form of streaming upload here for music and other large format forms. Multipart isn't the best option here I think.

kennylavender: @krobi64 do you know of anything out there that handles streams without multipart? or information on other ways? I couldn't find anything.

krobi64: Do we use an http server other than the library in node?

kennylavender: not at the moment.

src/API/FileController.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  private log = (logType: LogTypes) => (label: string) => <T extends any>(value?: T ): T => {
+    this.logger[logType]({ value }, label)
+    return value
+  }
+
+  private uploadStream = (stream: fs.ReadStream) => {
+    const formData = new FormData()
+    formData.append('file', stream)
+    return fetch(`${IPFS_URL}/api/v0/add`, {
```
kennylavender: Need to make IPFS_URL come from configuration.

src/API/FileController.ts
```diff
@@ -0,0 +1,79 @@
...
+import { inject, injectable } from 'inversify'
+import { Collection, Db } from 'mongodb'
+import fetch, {Response} from 'node-fetch'
+import * as Pino from 'pino'
+import { map } from 'ramda'
+
+import { childWithFileName } from 'Helpers/Logging'
+import { asyncPipe } from 'Helpers/asyncPipe'
+
+const IPFS_URL = process.env.IPFS_URL || 'http://localhost:5001'
```
kennylavender: Should come from configuration

src/API/FileController.ts
```diff
@@ -0,0 +1,79 @@
...
+
+interface FileResponseJson {
+  readonly Hash: string
```
kennylavender: I think we will want to use lowercase hash as the property we expose to users.

tests/integration/API/post-file.test.ts
```diff
@@ -0,0 +1,64 @@
...
+import { createIPFS } from '../../helpers/ipfs'
+import { runtimeId, setUpServerAndDb } from '../../helpers/utils'
+
+const PREFIX = `test-functional-node-poet-${runtimeId()}`
+const NODE_PORT = '28081'
+
+const ipfs = createIPFS()
+const fileHelper = createFileHelper({ port: NODE_PORT  })
+
+const getHash = prop('Hash')
```
kennylavender: `Hash` should be lowercase, `hash`. The code should convert this from IPFS response.

src/API/FileController.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  private handleFile = asyncPipe(
+    this.addFileToIPFS,
+    this.getResponseJson,
+    this.storeIPFSHash,
+  )
+
+  // using Promise.all directly in asyncPipe results in TypeError
+  // "Promise.all called on non-object"
+  private awaitAll = (xs: Array<Promise<any>>) => Promise.all(xs)
```
kennylavender: Maybe make this into a helper

src/API/FileController.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  private addFileToIPFS = (fileStream: fs.ReadStream) => this.uploadStream(fileStream)
+
+  private storeIPFSHash = async (response: FileResponseJson) => {
+    const { Hash: hash } = response
+    await this.collection.insertOne({ hash })
+    return response
+  }
+
+  private handleFile = asyncPipe(
```
kennylavender: add logs

kennylavender:
Have some minor things left to tackle, but would be good to get some more feedback at this point I think.

tests/integration/API/PostFile.test.ts
```diff
@@ -0,0 +1,64 @@
...
+describe('POST /files', async assert => {
+  // Setup Mongodb and the app server
+  const { db, server } = await setUpServerAndDb({ PREFIX, NODE_PORT })
+  {
+    const given = 'a single utf8 mardown file as a multipart file stream'
+
+    assert({
+      given,
+      should: 'return a hash that can be used to verify the file',
+      actual: await testMultiPartFileStreamResponseHash([files.utf8Markdown]).catch(identity),
```
kennylavender: Could use some feedback on these complete test functions, not sure if they are better than the normal way of calling the endpoint then testing the response with a bunch of assertions.

tests/integration/API/PostFile.test.ts
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const testStatusCode = asyncPipe(map(fs.createReadStream), fileHelper.postFileStreams, getStatus)
+
+describe('POST /files', async assert => {
+  // Setup Mongodb and the app server
+  const { db, server } = await setUpServerAndDb({ PREFIX, NODE_PORT })
+
+  {
+    const given = 'a single utf8 mardown file as a multipart file stream'
```
krobi64: `mardown` spelling?

tests/integration/API/PostFile.test.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/* tslint:disable-next-line:max-line-length */
+const testResponseHashes = asyncPipe(
+  map(fs.createReadStream),
+  fileHelper.postFileStreams,
+  (x) => x.json(),
+  map(getHash),
+  map(ipfs.fetchFileContent),
+  Promise.all.bind(Promise),
```
krobi64: why not use `pipeP`?

src/API/FileController.ts
```diff
@@ -0,0 +1,94 @@
...
+
+interface FileResponseJson {
+  readonly hash: string
+}
+
+interface File {
+  readonly path: string
+}
+
+enum LogTypes {
```
krobi64: This feels out of place in the FileController. Can you put it in Interfaces?

src/API/FileController.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  private addFileToIPFS = (fileStream: fs.ReadStream) => this.uploadStream(fileStream)
+
+  private storeIPFSHash = async (response: FileResponseJson) => {
+    const { hash } = response
+    await this.collection.insertOne({ hash })
+    return response
+  }
+
+  private handleFile = asyncPipe(
```
kennylavender: Not sure we need all these log traces in this async pipe. Logs in the main might be sufficient.

kennylavender:
@krobi64 sorry, i didn't mean to dismiss this new review, I thought i was dismissing the previous one lol

tests/integration/API/PostFile.test.ts
kennylavender: pipeP only works on async functions, we created asyncPipe using async/await it works with async and sync functions.

tests/integration/API/PostFile.test.ts
krobi64: took me a while to understand what you were doing. It doesn't call out a specific error as readily. It does make it easier to write. What is the `.catch(identity)`? Seems like all you would catch are errors?

src/API/FileController.ts
kennylavender: Yeah, that might be another good spot. I plan on refactoring this at some point, I am still experimenting a bit with this and the log method. I might be creating a method creator helper, so I was thinking about sticking the types with the method helper

tests/integration/API/PostFile.test.ts
krobi64: gotcha.

tests/integration/API/PostFile.test.ts
kennylavender: Yeah its catching any errors so the test does not crash the test runner. Its sort of taking place of riteways `Try` at the moment because riteways `Try` doesn't work for async functions. I've created a pr to adjust `Try` https://github.com/ericelliott/riteway/pull/18

kennylavender: > It doesn't call out a specific error as readily.

Yeah its not testing for errors, this assertion is only checking that the returned hashes result in the correct and full file from ipfs. No partial files, etc.

kennylavender: I made the tests this way based on your comment above https://github.com/poetapp/node/pull/612/files/0b45faca2c50b636979c9bb9fd64270a551f3fa8#diff-44e42c1f1cb5e71ade2c954f14c5cd3d

I may have taken it too far though

kennylavender:
Just realized an issue while updating the README. Because we allow multiple uploads we need to return some sort of identifier with each upload. We can use the name of the file and/or the fieldname.

kennylavender:
I think tackling multiple uploads and handling fields and file names can bed pushed off to the future by limiting the fields and uploads to 1. I've made that change.

kennylavender:
@krobi64 added the update to the readme.

README.md
```diff
@@ -153,7 +153,7 @@ Example:
 
 ```
 [
```
krobi64: We're returning an array because the endpoint supports multiple files? We should modify line 148 to indicate it can be multiple files.

> Takes a multipart file upload. Only accepts 1 field which is a file.

kennylavender: We are returning an array because we can support multiple files in the future, right via this pr it only supports 1.

kennylavender: Adjusted the test, maybe the new text is more clear?

src/API/FileController.ts
lautarodragan: The strategy for now is too much is better than too little, regarding logs. We can downsize after we're in production and know everything's going well.

src/API/FileController.ts
```diff
@@ -0,0 +1,95 @@
...
+
+interface IPFSFileResponseJson {
+  readonly Hash: string
+}
+
+interface FileResponseJson {
+  readonly hash: string
+}
+
+interface File {
```
lautarodragan: This guy is marked as unused by my IDE (but it could be lying)

kennylavender: Yup, looks like I forgot to delete it. Shouldn't the linter be catching this when ran on travis?

src/API/FileController.ts
```diff
@@ -0,0 +1,91 @@
...
+  info = 'info',
+  trace = 'trace',
+  error = 'error',
+}
+
+export interface FileControllerConfiguration {
+  readonly ipfsUrl: string
+}
+
+export const convertJson = (x: IPFSFileResponseJson) => ({
```
lautarodragan: Does this return a FileResponseJson?

kennylavender: Yup, thx, looks like I forgot to add the return type

src/API/Router.ts
```diff
@@ -90,6 +101,14 @@ export class Router {
     await this.server.close()
   }
 
+  private postFile = async (context: KoaRouter.IRouterContext, next: () => Promise<any>) => {
+    const files = context.request.files || {}
```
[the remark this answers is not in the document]
kennylavender: I think we should throw an error to give the user a heads up they didn't pass in a file or did not pass in a file correctly. I was thinking we should handle this a future pr as I wanted tests for that case and we have too much code in this PR already. What do you think?

src/API/Router.ts
```diff
@@ -66,7 +76,8 @@ export class Router {
         limit: Joi.number().optional(),
       },
     }
-
+    /* tslint:disable-next-line:max-line-length */
```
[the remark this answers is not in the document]
kennylavender: Let me try multi lines manually, I forgot we no longer have prettier running.

tests/integration/API/PostFile.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      given,
+      should: 'return a hash that can be used to verify the file',
+      actual: await testResponseHashes([files.utf8Markdown]).catch(identity),
+      expected: [fs.readFileSync(files.utf8Markdown).toString()],
+    })
+
+    assert({
+      given,
+      should: 'return the correct status',
+      actual: await testStatusCode([files.utf8Markdown]).catch(identity),
```
lautarodragan: Is there a need to catch? Won't errors get caught by tape and marked as a failed test?

tests/integration/API/PostFile.test.ts
```diff
@@ -0,0 +1,63 @@
+/* tslint:disable:no-relative-imports */
```
lautarodragan: This test file is not really testing the API/PostFile, is it?

tests/integration/API/PostFile.test.ts
kennylavender: tape will crash and not run the rest of the tests, its nicer when the error gets caught and lets the remaining tests show their output as well.

tests/integration/API/PostFile.test.ts
kennylavender: Not sure what you mean, this matches the other files in the folder

src/API/Router.ts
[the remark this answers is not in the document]
kennylavender: manual multi lines seems to have worked, removed the disable comments.

tests/integration/API/PostFile.test.ts
lautarodragan: got it, thanks

tests/integration/API/PostFile.test.ts
lautarodragan: Nevermind, my bad

tests/integration/API/PostFile.test.ts
lautarodragan: got it, thanks

src/API/Router.ts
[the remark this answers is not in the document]
lautarodragan: Sounds good!

src/API/FileController.ts
lautarodragan: I'd expect it to, but not sure. Raised https://github.com/poetapp/tslint-rules/issues/153, we can have a look in the future.

tests/integration/API/PostFile.test.ts
```diff
@@ -0,0 +1,63 @@
...
+const getHash = prop('hash')
+const getStatus = prop('status')
+
+const calcPath = (relPath: string) => path.resolve(__dirname, relPath)
+
+const files = {
+  utf8Markdown: calcPath('../../fixtures/files/utf8.md'),
+}
+
+/* tslint:disable-next-line:max-line-length */
```
lautarodragan: Is this disable needed?

kennylavender: I think this one was needed, but I will double check.

kennylavender: Nope, it wasn't needed, guess it was my imagination.

kennylavender:
Thanks!

PR #615, opened 2018-10-23
Wallet Balance Alert: Health

### LINKED ISSUE #435: Add Wallet Balance Alert to Health Endpoint

jwicks31:

### Release

## Description of Changes

Resolves #435 

- Added `addBalanceLowBool` to `refreshWalletInfo` in `src/Health/HealthController`
- Added `isBalanceLow` function with unit tests


src/Health/HealthController.ts
```diff
@@ -98,6 +107,13 @@ export class HealthController {
     return pickWalletInfoKeys(await this.bitcoinCore.getWalletInfo())
   }
 
+  private async checkWalletBalance(walletInfo: WalletInfo): Promise<WalletInfo> {
+    const { balance } = walletInfo
+    return balance < this.configuration.lowWalletBalanceBTC
```
kennylavender: Are we exposing the balance of the wallet? We probably shouldn't. The bool of balanceLow is probably all we should expose.

src/Health/HealthController.ts
```diff
@@ -98,6 +107,13 @@ export class HealthController {
     return pickWalletInfoKeys(await this.bitcoinCore.getWalletInfo())
   }
 
+  private async checkWalletBalance(walletInfo: WalletInfo): Promise<WalletInfo> {
```
kennylavender: This looks like a pure function that could be unit tested. Name could be better too, try to think through what this functions true responsibility is.

src/Health/HealthController.ts
lautarodragan: No no, it's ok to expose the balance of the wallet. The node won't be public in any environment for us and shouldn't for anyone else. 

None should set up a node directly against the public internet.

src/Health/HealthController.ts
jwicks31: Ok I updated the name and moved the checking logic to another function with unit tests. Looking better?

src/Health/HealthController.ts
kennylavender: Ok sounds good, thanks!

jwicks31:
@kennylavender should I go ahead and add the functional tests + integration in this PR? Or should I create an issue and do it separately?

kennylavender:
@jwicks31 tagging QE as it would be good to get their input on what they want you to do for the tests.

src/Health/HealthController.ts
```diff
@@ -98,6 +109,12 @@ export class HealthController {
     return pickWalletInfoKeys(await this.bitcoinCore.getWalletInfo())
   }
 
+  private async addBalanceLowBool(walletInfo: WalletInfo): Promise<WalletInfo> {
```
kennylavender: Looks like you got part of the logic unit tested, I think we have more testable logic though. I think we can test it all including the destructuring and adding a new property to a new object if we make a function that takes a wallet and a lowWalletAmount.

```ts
type addWalletIsBalanceLow = (lowBalanceAmount: number) => (wallet: WalletInfo) => WalletInfo
```

We can call that function from our Promise function that knows about the configuration.

jwicks31: Ok cool. I updated this.

src/Health/HealthController.test.ts
```diff
@@ -17,3 +17,41 @@ describe('isStatus200()', async (assert: any) => {
...
+      actual: addWalletIsBalanceBelow2(walletInfo),
+      expected: { ...walletInfo, isBalanceLow: false },
+    })
+  }
+
+  {
+    const walletInfo = walletGen(1)
+    assert({
+      given: 'a lowBalanceAmount higher than balance property of walletInfo',
+      should: 'return walletInfo with correct isBalanceLow bool',
```
kennylavender: make should a constant so it's not retyped in all 3 assertions.

PR #622, opened 2018-10-23
fix: update README

krobi64:

### Release

## Description of Changes

update README to reflect changes in node


README.md
```diff
@@ -181,46 +181,100 @@ A Po.et Claim is a JSON object that holds arbitrary information plus a few attri
...
+      'sec:nonce': 'cjn5czg1u0000mnc93s13ihuz',
+    },
+  },
 }
-
 ```
 
-The `publicKey` field must be set to the public part of a key pair you own. You'll need the corresponding private key to prove this claim was generated by you.
+The `issuer` field contains the public part of a key pair you own. You'll need the corresponding private key to prove 
+this claim was generated by you. Frost will generate an issuer for your account, or you can use `@poet/poet-js` to
```
kennylavender: Not sure if we are currently doing so elsewhere, but the node is meant as a base for other people to build from as well, so I don't think we should mention Frost.

krobi64: ack

PR #635, opened 2018-10-24
Functional Test for WalletInfo.isBalanceLow

### LINKED ISSUE #634: Add functional tests for /health endpoint isBalanceLow

jwicks31:

### Release

## Description of Changes

Resolves #634 

- Renamed Configuration Variable for `lowWalletBalanceInBitcoin` because `BTC` was made into `B_T_C`
- Copied/Modified `ensureBitcoinBalance` to be `ensureBitcoinBalanceAboveMin` and run recursively
- Added Test to ensure isBalanceLow reads false when the balance is above the `lowWalletBalanceInBitcoin` env variable.

We could have a test to ensure that it does say true if the balance is below it, but it would depend on the bitcoind instance, so could fail when running on local machines.


src/Configuration.ts
```diff
@@ -102,7 +102,7 @@ const defaultConfiguration: Configuration = {
   anchorIntervalInSeconds: 30,
 
   healthIntervalInSeconds: 30,
-  lowWalletBalanceBTC: 500,
+  lowWalletBalanceInBitcoin: 500,
```
warrenv: Where did this number com from? 
500 * ~6500USD/BTC = ~3.2 million USD

tests/helpers/bitcoin.ts
```diff
@@ -2,3 +2,10 @@ export const ensureBitcoinBalance = async (client: any, blocks: number = 101) =>
   const balance = await client.getBalance()
   if (balance === 0) await client.generate(blocks)
 }
+
+export const ensureBitcoinBalanceAboveMin = async (client: any, minBalance: number = 0, blocks: number = 101) => {
+  const balance = await client.getBalance()
+  if (balance < minBalance) await client.generate(blocks)
+  else return
+  await ensureBitcoinBalanceAboveMin(client, minBalance, blocks)
```
warrenv: Why is this recursively calling itself?

src/Configuration.ts
jwicks31: Sorry when I did the number I was thinking it was going to be in dollars and didn't change it after I realized it was BTC. I'll lower this.

tests/helpers/bitcoin.ts
jwicks31: I wasn't sure how else to ensure the number of blocks mined would raise the balance above the min requested.

warrenv: ok, but If for some reason the generate fails silently, it will basically be in an infinite loop, blow the stack and crash the program. At a minimum, there would need to be some way to safely stop the recursion. This is a tricky problem in the case of bitcoin because the block rewards keeps getting halved over time.

Since this is a test helper and we start with a fresh blockchain, the block reward is 25 btc/block. run it once. Generating 101 blocks in the tests ensures there's way more btc than even the current min of 500. 

This is tricky to correctly test right now because the blockchain is currently shared between tests. Unless this test is always the first one, it's unlikely that the balance would ever be zero and trigger a low balance alert. This will probably change soon so we may want to comment out these tests and mark them as TODO.

jwicks31: @warrenv How does it look now? Since we are sure that the `ensureBitcoinBalance` is going to generate 25*101 BTC and I have lowered the min to 1BTC is it OK to keep the: `should: 'return walletInfo property isBalanceLow as false',` test? I've added a commented out test before the `ensureBitcoinBalance` is called that will test for `should: 'return walletInfo property isBalanceLow as true',`

warrenv: Yes, that's looking much better!

PR #639, opened 2018-10-24
add to-promise

kennylavender:

### Release

## Description of Changes

I had another use case where I wanted to compose promises and sync functions, so I created helper so that we can compose together sync functions with pipeP

```ts
import { toPromise } from 'Helpers/to-promise'
const myFunc = pipeP(myPromise, toPromise(mySyncFunc), myOtherPromise)
```

I've included a few extra object tests just to make sure that types don't complain, I tried to make this type safe using generic types


lautarodragan:
@kennylavender can you try implementing the old pipeP you introduced instead?

using await in the pipe instead of then should allow us to pass both promise-returning functions and non-promise-returning functions to it, which would reduce boilerplate

kennylavender:
@lautarodragan Will do, this function may still be useful, for example I am going to use to create promise versions of sync functions to test asyncPipe with. Merging so I can finish up asyncPIpe now

PR #643, opened 2018-10-25
Implement asyncPipe

kennylavender:

### Release

## Description of Changes

I couldn't find a way to use better types than any. Let me know if you have ideas.

I am using .catch instead of RITEways Try because RITEways Try does not await promises and can still crash the test runner.


src/Helpers/async-pipe.ts
```diff
@@ -0,0 +1,4 @@
+type Func = (v?: any) => any | Promise<any>
```
lautarodragan: Sadly TypeScript doesn't have any decent support for the the most important FP functions like curry, compose, pipe... and even some cases of reduce. See this [comment by author of Ramda's type declarations](https://github.com/Microsoft/TypeScript/issues/10247#issuecomment-266653910).

You can hack around TS' limitation by providing one function overload per function arity.

It isn't pretty or very maintainable, but I think it still would add a lot of value since we rarely compose/pipe too many functions and it'd give us type safety in our pipes.

Wanna give it a shot?

Feel free to merge as-is though, code looks good and the tests are awesome.

kennylavender: Yeah, for some reason I though they had solved it partial in TS 3.0, but when I started looking around it seems they have not.

I think I will merge this as is, then I will go back and add the better typings.

Issue #644, opened 2018-10-25
Set Default Version to 1


lautarodragan:
Since we're introducing a bunch of breaking changes and decided not to be backwards compatible we can go ahead and use version 1.

Issue #608, opened 2018-10-22
Change startup defaults for blockchain and block height


geoffturk:
Default to testnet and configure production to use mainnet on the proper block height.
Also, use a more recent testnet block height.

PR #645, opened 2018-10-25
reset default version

lautarodragan:

### Release

## Description of Changes

Fixes https://github.com/poetapp/node/issues/644


ericelliott:
I don't have a good understanding of the motivation or impact of this change, or how it relates to #608. Is there a good explanation in one place?

lautarodragan:
@ericelliott it relates in that since we're going to ignore all blocks in the blockchain prior to the mainnet release (which is what the issue I link to is about), there won't be any version clash (the current stuff stored with version 1, 2, etc is never going to be read and is incompatible with the current node, which would just ignore it anyway).

Motivation: to avoid confusion. Even though we're using version 02, there are not races of previous versions anywhere in the code, nor any documentation about them.

PR #664, opened 2018-10-26
Improve AsyncPipe

kennylavender:

### Release

## Description of Changes

- better types
- fix implementation and types for no initial value
- had to switch to function declaration implementation for overloads to be accepted properly


src/Helpers/async-pipe.test.ts
```diff
@@ -31,4 +32,13 @@ describe('asyncPipe', async (assert: any) => {
       expected: await incP(double(inc(await doubleP(value)))),
     })
   }
+
+  {
+    assert({
+      given: 'no starting value',
+      should,
+      actual: await asyncPipe(fixedNum, incP)().catch(identity),
```
lautarodragan: Is the catch here needed?

kennylavender: Not really for this particular test, we do need a test for it though to make sure the types don't complain, I will remove the catch from all the current tests and make a specific test.

src/Helpers/async-pipe.ts
```diff
@@ -1,4 +1,86 @@
-type Func = (v?: any) => any | Promise<any>
+export function asyncPipe<V0 extends any, T1>(fn1: (x?: V0) => T1 | Promise<T1>): (x?: V0) => Promise<T1>
```
lautarodragan: Use `asyncPipe` as a name (following AirBnB's naming standard for files)

kennylavender: will rename in follow up PR 👍

PR #697, opened 2018-10-31
fix(BatchReader): test response object in getDirectoryFileHashes

### LINKED ISSUE #650: Bug: Read next directory failure
```
[..] ERROR (..): Read next directory failure
    module: "BatchReader"
    file: "Router"
    method: "onBatchReaderReadNextDirectoryRequest"
    error: {
      "stack": [
        "TypeError: Cannot read property '9tmuBnNa7juruX4fFjWMBUQ698PownJ3yrZj2modX7KjQ32wB4uWGrm3tPCeURu' of undefined",
        "    at IPFS.getDirectoryFileHashes (.../node/dist/babel/src/BatchReader/IPFS.js:30:36)",
      ],
      "message": "Cannot read property '9tmuBnNa7juruX4fFjWMBUQ698PownJ3yrZj2modX7KjQ32wB4uWGrm3tPCeURu' of undefined",
      "type": "TypeError"
    }

```

```js
> db.batchReader.find({ipfsDirectoryHash: '9tmuBnNa7juruX4fFjWMBUQ698PownJ3yrZj2modX7KjQ32wB4uWGrm3tPCeURu'}).pretty()
{
	"_id" : ObjectId("5bd160e2c3c8ec092d3f256b"),
	"ipfsDirectoryHash" : "9tmuBnNa7juruX4fFjWMBUQ698PownJ3yrZj2modX7KjQ32wB4uWGrm3tPCeURu",
	"ipfsFileHashes" : null,
	"lastAttemptTime" : 1540448877738,
	"successTime" : null,
	"attempts" : 1
}
```

```
$ curl http://localhost:5001/api/v0/file/ls?arg=9tmuBnNa7juruX4fFjWMBUQ698PownJ3yrZj2modX7KjQ32wB4uWGrm3tPCeURu
{"Message":"invalid 'ipfs ref' path","Code":0,"Type":"error"}
```

https://github.com/poetapp/node/blob/67c76843da35dd0e93fe6e2f9cce0002644b11e1/src/BatchReader/IPFS.ts#L49-L52

I'm not sure why that ipfsDirectoryHash is causing an error, but `getDirectoryFileHashes` seems to be assuming `response.Objects` will never be `undefined`.

Finding `ipfsDirectoryhash`es that point nowhere in the blockchain is normal behavior and we should just ignore these.

krobi64:

### Release

## Description of Changes

Fixes #650

Put a check in `getDirectoryFileHashes` to ensure the response has the necessary components. Return an empty array if it doesn't


src/Helpers/isNotNil.test.ts
```diff
@@ -0,0 +1,51 @@
+import { describe } from 'riteway'
+
+import { isNotNil } from './isNotNil'
+
+describe('isNotNil', async assert => {
+
+  const returnTrue = 'return true'
```
kennylavender: Might just make this a general should, `return the correct boolean` so its not duplicate of the expected

lautarodragan: My 2 cents: I find it more readable to have the should right there, and to tell me exactly what it should be doing.

"Return the correct boolean" reads to me almost as "it should be working well and the return type should be a boolean"

src/BatchReader/IPFS.ts
```diff
@@ -48,7 +50,10 @@ export class IPFS {
 
   getDirectoryFileHashes: getDirectoryFileHashes = async (hash: string) => {
     const response = await this.ls(hash)
-    return response.Objects[hash].Links.map(x => x.Hash)
+    const validResponse = isNotNil(response.Objects) &&
```
lautarodragan: Dunno if needed, but maybe add isNotNil(response) just in case? or response will never be undefined or a primitive?

src/Helpers/isNotNil.test.ts
kennylavender: I used to do it that way, but I found that it created tests where the expected/should became out of sync over time. This is a very simple function so it's not likely but for more complicated functions where you might adjust/change/fix/addto the function and tests overtime they easily become out of sync and can cause confusion when reading the test results.

lautarodragan: I hear you, but wouldn't the added responsibility of keeping these should/given up to date correctly be part of the philosophy of RITEway?

I used to prefer the test code being the single source of truth and avoiding human-readable messages for the very same reason, but I'm not sure now, since the RITEway does feel like better documentation of purposes...

krobi64: R = Readable. I do believe that being explicit about the expected return in the should documents the test cases better. Remember, that reading the test output is also a way to see expected behavior:

```
ok 98 Given an empty string: should return true
ok 99 Given an empty array: should return true
ok 100 Given an empty object: should return true
ok 101 Given undefined: should return false
ok 102 Given null: should return false
ok 103 Given no value: should return false
```
I believe that more clearly documents the expected behavior than `return the correct boolean`

kennylavender: Commenting on all of this here:

> I hear you, but wouldn't the added responsibility of keeping these should/given up to date correctly be part of the philosophy of RITEway?

> Readable. I do believe that being explicit about the expected return in the should documents the test cases better. Remember, that reading the test output is also a way to see expected behavior

> ok 98 Given an empty string: should return true
> ok 99 Given an empty array: should return true
> ok 100 Given an empty object: should return true
> ok 101 Given undefined: should return false
> ok 102 Given null: should return false
> ok 103 Given no value: should return false

How often do you actually read the output of all the passed tests? I don't ever, I mean they look nice and all when they pass but I ignore them. :)

Eric was the one that pointed it out to me to not duplicate the actual/expected in the given/should ( I don't believe he meant as a hard rule or anything, just a suggestion )

Its the failing tests we care about and that should answer the 5 questions, and those display the actual/expected in addition to given/should when they fails so I don't see a reason to duplicate the expected in the should. 

I don't think its right or wrong either way in my opinion, just one is easier to forget to update during maintenance. Feel free to duplicate the expected in the should if you would like, I won't complain I was just offering a suggestion :)

kennylavender: its also possible that things like booleans and null, undefined are just too low level not complicated enough to not duplicate, so you both could very well be right and its better to name them in the given/should

krobi64: fwiw: on new libraries, I do look at the results of the tests to get a quick read of expected behavior. Once I'm familiar, then yeah, I just look at the failing tests.

I don't have an ultra strong opinion, but a mild preference. 😀

kennylavender: > I don't have an ultra strong opinion, but a mild preference.

Same, I am willing to try out the more duplicated way if we want to do that as a team. :D

lautarodragan: Same here. Let's try that as a team and see how it goes.

kennylavender: Sounds good, maybe il be proved wrong :D

Issue #177, opened 2018-06-23
IPFS Upload Integrity Check — add + cat


lautarodragan:
needed-by #53 
needed-by #112 

# Tasks

- [x] RDD
- [x] There are ~20% broken files in production. Find a way to reproduce the bug locally, achieving similar results. We'll need a script that we can run to massively post files to the node.
- [x] Ideally, find a way to reproduce the bug for a single claim deterministically.
- [x] Break IPFS Upload out of Storage module, separating Read and Write operations into two modules: `StorageReader` and `StorageWriter`.
- [x] Make IFPS Upload async, similar to our current IPFS Download process.
- [ ] Add integrity check to `StorageWriter`, storing the result of the check in the database.

kennylavender:
Blocked by #396

PR #690, opened 2018-10-31
IPFS Integrity Check

kennylavender:
[PR Process](https://github.com/poetapp/documentation/blob/master/process/pr-review.md#pr-process) - [PR Review Checklist](https://github.com/poetapp/documentation/blob/master/process/pr-review.md#pr-review-checklist)

### Release
 Semantic release is enabled for this repository. Make sure you follow the right commit message convention. 
 We're using semantic-release's default — [Angular Commit Message Conventions](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines).

## Description of Changes

closes #177 

Add integrity check step to the claim controller. The point is so that we can be notified if a claim was not correctly uploaded to IPFS and record the information so we can have a better reason as to why the claim was not correctly uploaded IPFS.

### Make it possible for claims to break
Adjust the IPFS helper so it will break with non-ascii characters. Change this line

https://github.com/poetapp/node/blob/552b4b81acb778f9cf32d8723e7922e381e0b978/src/Helpers/IPFS.ts#L40
to
```
knownLength: text.length,
```
### Adjust your docker-compose.override.yml file
```yml
version: '3.6'

services:
  poet-node:
    volumes:
      - .:/usr/src/app
    command: sleep infinity
```

#### Run the node and requirements
* term1: `docker-compose up -d`
* term1: `docker-compose exec poet-node bash`
* term1: `npm run build && npm run start`

#### Post a working claim

* term2 `npm run create-claim` then wait for the log that the claim was stored succesfully

#### Post a claim that will trigger the Integrity check to fail
Create a claim with non-ascii characters that will fail due to our purpose built failure in the IPFS helper. You should see that a IntegrityCheckFailure was thrown. The name we pass create-claim needs to be long enough to cause a length mismatch for the stream

[...]

PR #710, opened 2018-11-01
Implement IPFS Helper

kennylavender:

### Release

## Description of Changes

Extract just enough code as a helper so we can implement #177 without duplicating ipfs logic again. We can build on this in future prs.

I moved around the claim data file so its accessible to both sets of integration tests.
The integration tests are duplicates of `tests/integration/StorageWriter/IPFS.test.ts`, but I did not delete them because we need them to remain in place until we refactor the StorageWriter to use this helper.

I did do a little factor around the fetch options so that way users can provide a custom timeout.

## Out of scope
- refactoring modules to use this helper
- new integration tests
- refactoring out logic / unit tests
- adjusting logic


src/Helpers/IPFS.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const IPFS = ({
+  url = 'http://localhost:5001',
+}: IPFSConfiguration) => {
+  const paths = {
+    add: `${url}/api/v0/add`,
+    cat: `${url}/api/v0/cat`,
+  }
+
+  const addText = ({
```
krobi64: why `addText` and not `addContent`?

src/Helpers/IPFS.ts
```diff
@@ -0,0 +1,61 @@
...
+    timeout = minutesToMiliseconds(10),
+  }: FetchOptions = {}) => async (text: string): Promise<string> => {
+    const formData = new FormData()
+
+    formData.append('file', str(text), {
```
krobi64: Are we no longer using an input stream but reading everything into memory?

src/Helpers/IPFS.ts
```diff
@@ -0,0 +1,61 @@
...
+      knownLength: Buffer.from(text).length,
+      filename: 'file',
+      contentType: 'plain/text',
```
krobi64: why are we forcing this to `plain/text`?

src/Helpers/IPFS.ts
lautarodragan: I believe this is taken from the old use (which is still in place, outside the File API).

krobi64: Can one of you refer me to the code this is going to eventually replace? I don't know where this is supposed to be used then.

src/Helpers/IPFS.ts
kennylavender: This is old code, we are just moving it to a helper

kennylavender: https://github.com/poetapp/node/blob/master/src/StorageWriter/IPFS.ts

src/Helpers/IPFS.ts
kennylavender: https://github.com/poetapp/node/blob/master/src/StorageWriter/IPFS.ts
and
https://github.com/poetapp/node/blob/master/src/StorageReader/IPFS.ts

src/Helpers/IPFS.ts
kennylavender: Not sure if it is necessary, I believe its because this function was created to just add any text. Also out of scope of this PR

https://github.com/poetapp/node/blob/master/src/StorageWriter/IPFS.ts
and
https://github.com/poetapp/node/blob/master/src/StorageReader/IPFS.ts

src/Helpers/IPFS.ts
kennylavender: also in the middle of duplicating `cat` again
https://github.com/poetapp/node/pull/690 which is why we decided to extract the helper.

kennylavender:
> But I don't think there's any case for choosing a different timeout for each call to the same function
I agree probably won't need to do this

> and it does lead to the awkward add()().
The more I do more and more functional programming with data last functions, the more normal this becomes ;)

I do think it feels extremely awkward to specialize the individual functions from the factory, its a pretty normal pattern to have the factory take certain config then the returned functions take more specialized config.
* https://github.com/ipfs/js-ipfs-api
* https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#filesadd

If its an issue with the fp style data last format `add = (config? FetchConfig)(text: string)` that feels awkward maybe we can just make the config a second optional argument? `add = (text: string, config?: FetchConfig)`

l don't mind going with whichever of 3 the versions you choose though.

src/Helpers/IPFS.ts
krobi64: gotcha.

src/Helpers/IPFS.ts
krobi64: ack

PR #715, opened 2018-11-02
Return void from DAO methods

jwicks31:

### Release

## Description of Changes

Changes DAO method types to return void


src/Health/HealthDAO.ts
```diff
@@ -33,15 +33,15 @@ export interface IPFSInfo {
   readonly ipfsIsConnected: boolean
 }
 
-type updateBlockchainInfo = (x: BlockchainInfo) => Promise<UpdateWriteOpResult>
+type updateBlockchainInfo = (x: BlockchainInfo) => void
```
kennylavender: should be Promise<void> I think? wonder why ts is complaining

lautarodragan: try, Promise<void>

kennylavender: yup that is what I meant, man... my brain today.

lautarodragan: Right, TS is complaining because we're trying to tell it "this returns a Promise<void>" but it knows better. Maybe a type cast? `return ... as Promise<void>`

PR #720, opened 2018-11-03
Add/Export types that represent the returned object and creator

kennylavender:

### Release

## Description of Changes

I forgot we need to export a separate interface when using factory functions because typescript doesn't infer the interface as it does with class.

See #721 for it in use.

- create/export IPFS interface
- separate types from the functions so they can be used for the IPFS interface.
- renamed factory from IPFS to createIPFS
- renamed local function in test file to avoid name collision.


src/Helpers/IPFS.ts
```diff
@@ -7,25 +7,35 @@ import { minutesToMiliseconds } from './Time'
...
+
+export interface IPFS {
+  addText: addText,
+  cat: cat
 }
 
-export const IPFS = ({
+type createIPFS = (config?: IPFSConfiguration) => IPFS
+
+export const createIPFS: createIPFS = ({
```
kennylavender: It seems I could have left this named `IPFS`, allowing IPFS to represent both the returned object and the factory.

```suggestion
export const IPFS: createIPFS = ({
```

TypeScript seemed to have no issues with that and it worked fine, but I couldn't find any TS documentation on it so I was unsure if it was just a bug.

lautarodragan: > It seems I could have left the factory named IPFS, allowing IPFS to represent both the returned object and the factory. 

Yep, that's a common pattern in TS. TS things and JS things can share the same name since TS things vanish after the build step, roughly speaking. A const and an interface having the same name isn't an issue.

I liked the idea of naming the factories after their returned object in PascalCase.

Maybe we can avoid having a function `IPFS` of type `createIPFS` too, by skipping the separate type declaration:

```ts
export const IPFS = ({
  url = 'http://localhost:5001',
}: IPFSConfiguration = {}): IPFS => {
```

Sometimes separating the type declaration from the implementation doesn't add that much value IMO.

I'm fine either way though.

PR #735, opened 2018-11-06
Blockchain Writer - Store the date of transaction creation

### LINKED ISSUE #734: Blockchain Writer - Store the date of transaction creation.
parent issue: #707 

BlockchainWriter will need to store the current date along with the transactionId as `transactionCreationDate`

kennylavender:

### Release

## Description of Changes

closes #734


src/BlockchainWriter/Controller.ts
```diff
@@ -87,7 +87,7 @@ export class Controller {
     const data = poetAnchorToData(poetAnchor)
     const txId = await anchorData(data)
 
-    await dao.setTransactionId(ipfsDirectoryHash, txId)
+    await dao.setTransactionId(ipfsDirectoryHash, txId, new Date().toISOString())
```
warrenv: Why pass the third argument when the function provides the same value as a default?

kennylavender: It might be argued that the controller is in control of the data and should be responsible for passing in a date even if the dao provided a default. IDC either way, just thought it was good to be explicit in this case.

warrenv: Then the default should be removed so it's explicit in every case right?

kennylavender: I wouldn't say `should`. If this were JavaScript I would have used a valid default to provide a type hint. In TS I could just make it a required type of string, but I still prefer valid defaults.

kennylavender: In TS I guess requiring a type of string would act more like argument validation in a limited way, but I am not really a fan of that, I would rather add real validation checking.

kennylavender: But I don't mind changing this up if you and the other prefer a change to this.

kennylavender: @lautarodragan you wanted to use an ISO string for the date correct?

kennylavender: actually, I see why this is a bit awakard after thinking about it a bit more, the dao function should be a bit more general and should know the least amount posssible as about the controller requirements, so I should probably update and rename the function to take an updated entry. `findAndUpdateByIPFSDirectoryHash` would make more sense, that allows the controller just to pass updates in as an object.

kennylavender: @warrenv I've updated it, let me know what you think.

warrenv: Yes, that makes more sense now.

src/BlockchainWriter/DAO.ts
```diff
@@ -1,6 +1,12 @@
 import { inject, injectable } from 'inversify'
 import { Collection } from 'mongodb'
 
+interface Entry {
+  ipfsDirectoryHash: string
```
lautarodragan: readonly?

kennylavender: updated

src/BlockchainWriter/DAO.ts
```diff
@@ -17,10 +23,15 @@ export class DAO {
...
       ipfsDirectoryHash,
       txId: null,
+      transactionCreationDate: null,
     })
 
   readonly findTransactionlessEntry = () => this.blockchainWriterCollection.findOne({ txId: null })
 
-  readonly setTransactionId = (ipfsDirectoryHash: string, txId: string) =>
-    this.blockchainWriterCollection.updateOne({ ipfsDirectoryHash }, { $set: { txId } }, { upsert: true })
+  readonly findAndUpdateByIPFSDirectoryHash = ({ ipfsDirectoryHash, ...updates }: Entry) =>
```
kennylavender: This should return void.

src/BlockchainWriter/DAO.ts
```diff
@@ -17,10 +23,16 @@ export class DAO {
...
       ipfsDirectoryHash,
       txId: null,
+      transactionCreationDate: null,
     })
 
   readonly findTransactionlessEntry = () => this.blockchainWriterCollection.findOne({ txId: null })
 
-  readonly setTransactionId = (ipfsDirectoryHash: string, txId: string) =>
-    this.blockchainWriterCollection.updateOne({ ipfsDirectoryHash }, { $set: { txId } }, { upsert: true })
+  readonly findAndUpdateByIPFSDirectoryHash = async ({ ipfsDirectoryHash, ...updates }: Entry) => {
```
lautarodragan: Shouldn't this be named just `updateBy...` or did you mean to use [findAndModify](https://docs.mongodb.com/manual/reference/method/db.collection.findAndModify/)?

kennylavender: `updateByIPFSDirectoryHash` works for me, I think mongo func names are starting to invade my brain a bit.

PR #754, opened 2018-11-09
test: K6 Stress Testing Tool

wzalazar:

### Release

## Description of Changes

Fix https://github.com/poetapp/node/issues/755

This PR introduces:

- Script Endpoint to create Claims
- Test example, create `works`
- Documentation

This is an approach with `k6` https://docs.k6.io/docs.  

**Why is so difficult to create `works`?**

The endpoint works need a valid and unique claim for that, you must to use the library `poet-js`. For this reason, you need to create the claim first for use it after.

**why you do not create the all claim before and save it a file?**

it a possibility, but the problem is when you want to test for example concurrency with interval times. In this cases, you don't know how many claims you needed. Another problem is how to knows what claim you have to send.

Other libraries:

I researched `Artillery` and `JMeter`, and with these libraries is more complicated

**Steps to run the script:**

- Install K6 https://docs.k6.io/docs/installation
- `npm run create-claim-api`
- in the folder `tests/stress` `k6 run --vus 10 --duration 30s script.js`


docker-compose.yml
```diff
@@ -72,6 +72,17 @@ services:
       -rpcallowip=::/0
       -deprecatedrpc=signrawtransaction
 
+  k6:
+    build: ./tests/stress
+    depends_on:
+      - poet-node
+    links:
```
warrenv: links are no longer required. If you remove the links array, you should be able to docker-compose exec into the k6 container and ping poet-node.

wzalazar: awesome, I will test

wzalazar:
I tested in local and against an external domain and both works fine

for an external domain, I used https://regtest.qa.poetnetwork.net/works the last work is mine :)
or you can check if there is a work with

```
claim: {
  name: "testing claim 1542142376190",
  author: "the tester"
}
```

tests/stress/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:10.12.0
```
geoffturk: Change to 10.13.0

wzalazar: cool, thanks @geoffturk !!!!

warrenv: This was already done on master yesterday.

wzalazar:
@warrenv I applied the same refactoring I did on the Frost api

PR #771, opened 2018-11-14
fix: adjust sleep to fix rabbitmq connection issue

kennylavender:

### Release

## Description of Changes

Update node sleep time to give rabbitmq more time to startup.


docker-compose.yml
```diff
@@ -101,7 +101,7 @@ services:
       - ipfs
     command: >
```
warrenv: Even better would be to replace this with the wait-for-it script as is done in frost-api. :-)

lautarodragan: That would be nice, but maybe merge this one as is and use wait-for-it in another PR if this unblocks something...?

warrenv: If someone is blocked, they should use a `docker-compose.override.yml` to adjust the command for their particular situation.

PR #784, opened 2018-11-16
feature: /file endpoint, 400 if no file is provided

### LINKED ISSUE #699: API POST /files: Handle no files from multipart form data
We should probably throw an error when no files are found in the multipart form data to alert that user that they did not provide a file and/or they provided the file incorrectly. Include integration tests.

kennylavender:

### Release

## Description of Changes

resolves #699 

Fixed our exception middleware so that we can use `context.throw`.

Experimented with a different test format. I think I like this format better than the format above. This new format does the side effects and catches the errors then just uses pure functions to test the response and response json.


src/API/Router.ts
```diff
@@ -106,7 +106,12 @@ export class Router {
   }
 
   private postFile = async (context: KoaRouter.IRouterContext, next: () => Promise<any>) => {
+
     const files = context.request.files || {}
+
+    if (values(files).length <= 0)
+      context.throw(400, 'No file found.')
```
kennylavender: Was thinking about unit testing this logic.

src/API/Middlewares/HttpExceptionsMiddleware.ts
```diff
@@ -14,10 +14,13 @@ export const HttpExceptionsMiddleware: Koa.Middleware = async (context: any, nex
...
-        context.logger.error({ exception }, 'Error Caught at Middleware - Will Return Internal Server Error')
+        context.logger.error({ exception }, 'Error Caught at Middleware')
 
-      context.body = 'Internal Server Error'
-      context.status = 503
+      context.body = {
+        message: exception.message || 'Internal Server Error',
+      }
+      context.status = exception.status || 503
+      context.app.emit('error', exception, context)
```
kennylavender: Added this because its recommended by koa: https://github.com/koajs/koa/wiki/Error-Handling#catching-downstream-errors

src/API/Middlewares/HttpExceptionsMiddleware.ts
```diff
@@ -14,10 +14,13 @@ export const HttpExceptionsMiddleware: Koa.Middleware = async (context: any, nex
       context.status = 404
     } else {
       if (context.logger)
-        context.logger.error({ exception }, 'Error Caught at Middleware - Will Return Internal Server Error')
+        context.logger.error({ exception }, 'Error Caught at Middleware')
 
-      context.body = 'Internal Server Error'
-      context.status = 503
+      context.body = {
+        message: exception.message || 'Internal Server Error',
```
kennylavender: Instead of just returning a string for the body, I think it's better we return json with a message property, that allows us to all return additional details in the future as well.

krobi64: It takes us away from using `response.text()` but I do prefer structured errors returned.

kennylavender:
@krobi64 we don't have any documentation on the error formats I don't believe, but it is a good point. I reverted this to just create a text error for the time being, we can switch to formatted in a separate issue/pr.

src/API/Middlewares/HttpExceptionsMiddleware.ts
```diff
@@ -14,10 +14,11 @@ export const HttpExceptionsMiddleware: Koa.Middleware = async (context: any, nex
       context.status = 404
     } else {
       if (context.logger)
-        context.logger.error({ exception }, 'Error Caught at Middleware - Will Return Internal Server Error')
+        context.logger.error({ exception }, 'Error Caught at Middleware')
 
-      context.body = 'Internal Server Error'
-      context.status = 503
+      context.body = exception.message || 'Internal Server Error'
```
lautarodragan: You don't want to leak error details to the world by default. We need to either keep it hardcoded to 'Internal Server Error' or add some logic and configuration options and leak details in testing/qa but not in prod.

src/API/Router.ts
lautarodragan: @kennylavender just realized, I think this should be a 422, not a plain 400.

kennylavender: @lautarodragan created an issue here: https://github.com/poetapp/node/issues/814

Issue #781, opened 2018-11-15
Create BlockchainWriter Transaction Max Age Service


kennylavender:
Parent #707 

Create a service to check for blockchainWriter collection entries that have a txId, don't have a blockHash and blockHeight and their transactionCreationDate is older than the transaction max age. Remove the txId from those entries and increment an attempts field.

kennylavender:
blocked By #782

Issue #707, opened 2018-11-01
Blockchain Confirmations: Transaction Retries (Naive)


lautarodragan:
## Goal

We need a way to track how _old_ a transaction is, and replace it by a new one if older than a configured amount.

## Reference

RDD: [Blockchain Confirmations](https://github.com/poetapp/documentation/blob/master/rdd/blockchain-confirmations.md#blockchain-confirmations)

Vague description and implementation tip in pseudo-js: [#BlockchainWriter](https://github.com/poetapp/documentation/blob/master/rdd/blockchain-confirmations.md#blockchainwriter)

Parent issue: https://github.com/poetapp/node/issues/46

## Implementation Details
* [x] `BlockchainWriter` will need to store the current date [along with the `transactionId`](https://github.com/poetapp/node/blob/9edaf9bf67a741bf34c0a9791b704fa3e3f5567a/src/BlockchainWriter/Controller.ts#L90).
* [x] `BlockchainWriter` will need to be aware of newly scanned blocks
  * [x] Update entries in the `blockchainWriter` collection where `txId in block.transactionIds` setting `blockHash` and `blockHeight`
  * [x] ~Upsert entries in the `blockchainWriter` collection, comparing their creation date against the block creation date. If the don't already have a `blockHeight` set and the date difference is greater than the configured amount, clear out `txId` and `transactionCreationDate` fields and increasing an `attempts` field.~ #781 Create a service to check for `blockchainWriter` collection entries that have a txId, don't have a blockHash and blockHeight and their `transactionCreationDate` is older than the transaction max age. Remove the txId from those entries and increment an `attempts` field.
*  ~The View module should update with confirmation information (`transactionCreationDate`), to be returned by the API `GET /works` in the `.timestamp` object.~
* [x] The Health module should update with retry information, grouping amount of attempts by amount of transactions. IOW, how many second attempts, how many third attempts, and so forth.

Clearing out the `txId` will have the next iteration of the service pick it up again, creating a new transaction and broadcasting it.

## New Configuration Options

- `purgeStaleTransactionsIntervalInSeconds`
- `maximumTransactionAgeInBlocks`

## Out of Scope

- Smarter mechanism for retries, like upping fee. 

## Related Work

[...]

PR #789, opened 2018-11-17
chore: add blockHeight at time of creating transaction Id

krobi64:

### Release

## Description of Changes

References #781 #707 

Add the blockHeight at time of adding a transactionId to the BlockWriter DAO. This will be used to determine stale transaction ids that have not yet been added to the block chain.


src/BlockchainWriter/Controller.ts
```diff
@@ -103,11 +103,13 @@ export class Controller {
 
     const data = poetAnchorToData(poetAnchor)
     const txId = await anchorData(data)
+    const blockInfo: any = await this.bitcoinCore.getBlockchainInfo()
```
lautarodragan: Can you add an interface to https://github.com/poetapp/node/blob/master/src/Helpers/Bitcoin.ts instead of using `any`?

geoffturk:
@krobi64 https://github.com/poetapp/node/issues/707#issuecomment-439854041

src/BlockchainWriter/Controller.ts
krobi64: Adding it to bitcoincore.d.ts actually since it defines the return as any.

PR #798, opened 2018-11-20
chore(BlockWriterDAO): purge stale transactions 

krobi64:

### Release

## Description of Changes

References #781, #707 

Given a creationBlockHeight outside the configured range, set txId to null


src/BlockchainWriter/DAO.ts
```diff
@@ -32,6 +32,18 @@ export class DAO {
       blockHeight: null,
     })
 
+  readonly purgeStaleTransactions = (currentBlockHeight: number, maxBlockHeightDelta: number) =>
+    this.blockchainWriterCollection.updateMany(
+      {
+        txId: { $ne: null },
+        blockHeight: null,
```
kennylavender: I am guessing $ne is not empty, can we use $ne for blockHeight as well?

krobi64: not equal

krobi64: we are looking for entries where the txId is not null and the blockHeight is null.

src/BlockchainWriter/DAO.ts
```diff
@@ -32,6 +32,18 @@ export class DAO {
...
+        creationBlockHeight: { $lt: currentBlockHeight - maxBlockHeightDelta },
```
kennylavender: I feel like this new method has too much knowledge of the logic.

I think this function should just take a block height to compare against instead of doing the math. I think the math should be left in the controller.

src/BlockchainWriter/DAO.ts
kennylavender: ah yeah ok.

src/BlockchainWriter/DAO.ts
krobi64: ack

PR #816, opened 2018-11-26
fix: Adjust logging in API middleware

### LINKED ISSUE #806: No file found incorrectly being logged as error
https://github.com/poetapp/node/blob/96eb4658332e9d221818e4f633d39b74389d53ce/src/API/Middlewares/HttpExceptionsMiddleware.ts#L9-L25

@kennylavender sorry I missed this in https://github.com/poetapp/node/pull/784, but the `else` branch here is for errors that we aren't expecting (like the mongoDB driver throwing due to a broken connection).

We need to change the middleware's code back to how it was, and use `IllegalArgumentException` for the "No Files" exception.

kennylavender:

### Release

## Description of Changes

resolves #806 

~Based on my suggestion in #806, I've moved the logging to the nested else. If you don't agree this is the right direction let me know.~

Update: It was decided to revert to throwing the IllegalArgumentException.


kennylavender:
> , but I fear the current approach leads to code duplication.

I think I see what you're saying here. If we don't handle this error globally via a middleware then we need to duplicate the logic in each handler and you want to avoid duplication.

However, handling this error globally is inflexible, and can make changes more difficult in the future. Each route handler should have control over this logic. We should aim to compose together our route handlers instead of handling things globally.

We can avoid duplication by moving that logic into its own module/function/middleware that we then compose into only the route handlers that need it. The logic would be adjusted to use context.throw instead of setting the body and status manually.

Example middleware, not sure if I would go with a middleware but it was the quickest example.

```ts
import * as Koa from 'koa'

import { IllegalArgumentException } from '@po.et/poet-js'

export const IllegalArgumentMiddleware: Koa.Middleware = async (context: any, next: () => Promise<any>) => {
  try {
    await next()
  } catch (exception) {
    if (exception instanceof IllegalArgumentException)
      context.throw(422, exception.message)
  }
}
```

> We should be able to get the same effect with the previous version of the middleware and changing context.throw(400, 'No file found.') to throw new IllegalArgumentException('No file found.').

This doesn't feel like a good solution, we are throwing an error from poetjs, I feel like only poetjs should be throwing this error. It's like importing errors from mongodb, then throwing those errors in our DAO's manually.

Issue #832, opened 2018-11-30
Add a Health Check for Anchor Retries


krobi64:
## Goal

We need a way to track how often we are resubmitting batches before they become part of the blockchain.

## Reference

RDD: [Blockchain Confirmations](https://github.com/poetapp/documentation/blob/master/rdd/blockchain-confirmations.md#blockchain-confirmations)

Parent Issue: #707 

## Implementation Details
* [x] Have the health module listen to `ipfsHashTxId` event
* [x] Upsert the ipfsHash and the current txn id in a separate collection internal to the Health module. If the action results in an insert, set attempts = 1. If the txn id has changed increment attempts by 1. Record timestamp of upsert
* [x] on getHealth, report the number of batches with attempts > 1, grouped by attempts. Store result in health collection.
* [x] listen for blockDownloaded, purge db entries with the txIds.

## Out of Scope

* Managing a potential race condition between blockchainWriter and blockchainReader and getHealthCheck.

PR #838, opened 2018-12-03
feat(Health): inject ipfsHashTxId into Health module

krobi64:

### Release

## Description of Changes

References #832 

Inject the IPFS_HASH_TX_ID event into the Health module
Create a consumer for the event


src/Messaging/Messages.ts
```diff
@@ -12,6 +12,11 @@ export interface LightBlock {
   readonly height: number
 }
 
+export interface HashTxId {
```
lautarodragan: `Hash` feels too generic here to me. Without context it could get confusing. Maybe rename to `IPFSDirectoryHashTransactionId`?

krobi64: lol of course

lautarodragan: You're already using `ipfsHashTxId` for the exchange name, so maybe just go with that in pascal case?

krobi64: yeah that's more consistent

PR #872, opened 2018-12-06
feature(API): add transactionAnchorInfo to health

krobi64:

### Release

## Description of Changes

References #832 

Add an `IPFSDirectoryHashDAO` to the API module and tie in the report to the `GET /health` endpoint

### DIFF
--- a/src/API/API.ts
+++ b/src/API/API.ts
@@ -82,6 +82,7 @@ export class API {
     const healthController = new HealthController({
       dependencies: {
         db: this.dbConnection,
+        ipfsDireactoryHashDAO: ipfsDirectoryHashDAO,
       },
     })
 
--- a/src/API/API.ts
+++ b/src/API/API.ts
@@ -10,6 +10,7 @@ import { ExchangeConfiguration } from './ExchangeConfiguration'
 import { FileController, FileControllerConfiguration } from './FileController'
 import * as FileDAO from './FileDAO'
 import { HealthController } from './HealthController'
+import * as IPFSDirectoryHashDAO from './IPFSDirectoryHashDAO'
 import { Router } from './Router'
 import { WorkController } from './WorkController'
 
@@ -28,6 +29,7 @@ export class API {
   private router: Router
   private messaging: Messaging
   private fileCollection: Collection
+  private ipfsDirectoryHashCollection: Collection
 
   constructor(configuration: APIConfiguration) {
     this.configuration = configuration
@@ -40,6 +42,7 @@ export class API {
     this.dbConnection = await this.mongoClient.db()
 
     this.fileCollection = this.dbConnection.collection('files')
+    this.ipfsDirectoryHashCollection = this.dbConnection.collection('ipfsDirectoryHashInfo')
 
     this.messaging = new Messaging(this.configuration.rabbitmqUrl, this.configuration.exchanges)
     await this.messaging.start()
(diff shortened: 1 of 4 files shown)


src/API/HealthController.ts
```diff
@@ -10,10 +13,16 @@ interface HealthObject {
   readonly networkInfo: object
   readonly estimatedSmartFeeInfo: object
   readonly ipfsRetryInfo: object
+  readonly transactionAnchorRetryInfo: AnchorRetryReport | Error
+}
+
+interface Error {
```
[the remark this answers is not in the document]
krobi64: that works

src/API/API.ts
```diff
@@ -10,6 +10,7 @@ import { ExchangeConfiguration } from './ExchangeConfiguration'
 import { FileController, FileControllerConfiguration } from './FileController'
 import * as FileDAO from './FileDAO'
 import { HealthController } from './HealthController'
+import * as IPFSDirectoryHashDAO from './IPFSDirectoryHashDAO'
```
lautarodragan: can we avoid namespacing the import here?

krobi64: I was following Kenny's example. I think since the Symbols are no longer required, I can go back to a normal import

kennylavender: I was just experimenting, sorry, didn't make for it to be taken as an example of how we are changing the code.

krobi64: Should experiments make it into production code? They become examples of how to do things.

krobi64:
Is #874 a blocker? I knew this would need to be resolved so that we can truly separate the modules. I was already thinking about using RMQ to resolve this. There will be a similar DAO in the Health module, with the difference being that it inserts and updates entries, whereas this DAO only reads.

Refactoring this is a larger task.

@geoffturk what is the priority?

src/API/API.ts
kennylavender: No, it was a mistake.

kennylavender:
@krobi64 I am good with this being merged before #874 if @lautarodragan is.

I think it might be helpful to move the DAO to the Health module. API can import the DAO from Health, that should give us an easy way to see that we have broken the Module boundaries.

src/API/IPFSDirectoryHashDAO.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  readonly getAnchorRetryHealth: getAnchorRetryHealth = async () => {
+    const cursorArray = await this.collection.aggregate([
+      { $match: { attempts: { $ne: 1 } } },
+      { $group: { _id: '$attempts', count: { $sum: 1} } },
+    ]).toArray()
+
+    if (isNil(cursorArray) || cursorArray.length === 0) return []
+
+    return cursorArray.map( (item: AnchorRetryDAOResult) => ({ attempts: item._id, count: item.count }))
```
kennylavender: Please break this line into multiples lines.

src/API/IPFSDirectoryHashDAO.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.collection = collection
+  }
+
+  readonly getAnchorRetryHealth: getAnchorRetryHealth = async () => {
+    const cursorArray = await this.collection.aggregate([
+      { $match: { attempts: { $ne: 1 } } },
+      { $group: { _id: '$attempts', count: { $sum: 1} } },
+    ]).toArray()
+
+    if (isNil(cursorArray) || cursorArray.length === 0) return []
```
kennylavender: is ` || cursorArray.length === 0` this part needed? It should already be an empty array I would think?

src/API/IPFSDirectoryHashDAO.ts
```diff
@@ -0,0 +1,45 @@
+import { AnchorRetryDAOResult, getAnchorRetryHealth } from 'Interfaces'
+import { Collection } from 'mongodb'
+import { isNil } from 'ramda'
+
+export interface Entry {
+  hash: string
```
lautarodragan: readonly?

krobi64:
Closing in favor of a slightly different approach

PR #873, opened 2018-12-06
chore(travis): set logging level to warn

lautarodragan:
Excessive log output in the travis output makes it harder to spot test failures.


tests/helpers/utils.ts
```diff
@@ -48,6 +48,7 @@ export const setUpServerAndDb = async ({
     MONGODB_USER: db.settings.tempDbUser,
     MONGODB_PASSWORD: db.settings.tempDbPassword,
     EXCHANGE_PREFIX: PREFIX,
+    LOGGING_LEVEL: 'warn',
```
warrenv: Doesn't this set the logging level to warn for local development also?
If this is really just to address travis, I'd prefer it be set in the env block in `.travis.yml`.

lautarodragan: Hmm wouldn't it only set it to warn for integration tests?

docker-compose.travis.yml
```diff
@@ -0,0 +1,6 @@
+version: '3.6'
+
+services:
+  poet-node:
+    environment:
+      - LOGGIN_LEVEL=warn
```
warrenv: mis-spelling.

tests/helpers/utils.ts
warrenv: yes. integration tests running in travis, locally or anywhere else. 

I'd rather not make developers go thru another config step - usually when you are developing you want full logging in case your new code breaks something. By setting this in `.travis.yml` it should get the desired behavior without introducing the complications a second docker-compose.yml file.

warrenv: All this Pr really needs is one line added to .travis.yml.

lautarodragan: Just tried that and didn't work :(

kennylavender: I actually prefer it when only errors log by default, maybe warnings too. The tests should tell us that something broke, not the logs. I realize though that we are still short on tests so that might be a good argument for leaving all the logs.

krobi64: The tests tell us something is broken, but not where. Until I put a log message in while I was trying to ensure the bitcoin nodes were at the correct height, I had no idea how long it could sometimes take. And it was even slower on Travis than my individual box.

From many other companies I've worked at, all I've seen happen is engineers just change the level of their logging. Is there a way from Travis' site to set the logging level for an individual run? That would make it easier to 

1. Set the default log level so that it is sparse (error/warning)
1. Allow an individual engineer to run it at a different debug level for information
1. Catch unnecessarily high log messages in the PR.

lautarodragan: > The tests tell us something is broken, but not where.

We should aim to solve this by having smaller scoped tests and more coverage whenever possible. 

> Is there a way from Travis' site to set the logging level for an individual run?

I think not. See https://docs.travis-ci.com/user/environment-variables/.

An ugly approach is to modify the docker-compose.travis.yml in your PR to whatever suits your needs and revert it after you've found what you need or just before merging... 

Also, we're still free to have it however we want in our local environments. This should affect travis alone.

PR #891, opened 2018-12-10
feat(Health): delete anchored transactions from ipfsDirectoryHash collection

krobi64:

### Release

## Description of Changes

References #832 

Delete any entries in the `Health:ipfsDirectoryHashCollection` that have been anchored


src/Health/Router.ts
```diff
@@ -67,4 +70,16 @@ export class Router {
       logger.error({ error }, 'Failed to update ipfsHashFailures on health')
     }
   }
+
+  blockDownloadedConsumer = async (blockDownloaded: BlockDownloaded): Promise<void> => {
+    const logger = this.logger.child({ method: 'blockDownloadedConsumer' })
+
+    logger.trace({ blockDownloaded }, 'Block downloaded, removing associated transactions')
+    try {
+      const transactionIds = pluck('transactionId', blockDownloaded.poetBlockAnchors)
```
lautarodragan: Can the same be achieved with `blockDownloaded.poetBlockAnchors.map(_ => _.transactionId)`? Referencing fields by strings is not type-safe nor refactor/rename friendly.

krobi64: done. I didn't take the commit suggestion because I needed to remove the import as well

src/Health/HealthController.ts
```diff
@@ -167,6 +173,12 @@ export class HealthController {
     )
   }
 
+  public purgeIpfsDirectoryHashByTransactionIds = pipeP(
+    this.log(LogTypes.trace)('purging IPFSDirectoryHash for transactionIds'),
+    this.removeIPFSDirectoryHashByTransactionId,
```
lautarodragan: Why not directly do `this.ipfsDirectoryHashDAO.deleteByTransactionIds` here instead?

krobi64: I'm following a pattern in the controller to log elements. Are we changing the pattern?

src/Health/Router.ts
```diff
@@ -67,4 +69,16 @@ export class Router {
       logger.error({ error }, 'Failed to update ipfsHashFailures on health')
     }
   }
+
+  blockDownloadedConsumer = async (blockDownloaded: BlockDownloaded): Promise<void> => {
+    const logger = this.logger.child({ method: 'blockDownloadedConsumer' })
+
+    logger.trace({ blockDownloaded }, 'Block downloaded, removing associated transactions')
```
lautarodragan: Can you make this one a `debug` instead of `trace`?

krobi64: yep

src/Health/HealthController.ts
lautarodragan: Got it. Let's not

PR #929, opened 2018-12-14
refactor(BlockchainWriter): able to run as independent module

### LINKED ISSUE #868: Microservices Refactor: BlockchainWriter (POC)
Follow steps 1 through 3 of the [RDD](https://github.com/poetapp/documentation/blob/master/rdd/node-microservices.md#poet-node-microservices-refactor) for BlockchainWriter only.

Make the BlockchainWriter service ignore the `enableAnchoring` configuration option. It doesn't need to support it, and it will help us test this.

## Testing

Set `enableAnchoring` to `false` in the configuration, hard-code it to `true` in `src/index.ts` for BatchWriter, and start both the traditional non-microservice node and the new separate BlockchainWriter service. 

This will leave us in a state in which we basically have the full node running as usual, but with the BlockchainWriter service running separately.

We'll need to manually test this to ensure everything works as usual. Integration and functional tests may need to be adjusted to cover this, but let's leave that outside the scope of this issue.

krobi64:

### Release

## Description of Changes

Resolves #868 

1. Always run BatchWriter as if `enableAnchoring` is true
1. Add helpers for startup and shutdown
1. Add an index file to BlockchainWriter module
1. Add a script to `package.json` to run the index file
1. Add a Dockerfile with the specific command to run
1. Add a poet-node_blockchaIn-writer block to `docker-compose.yml`

TODO:
* [ ] Verify tests still run
* [ ] Determine a manual test
* [ ] See if there's a simple strategy to update tests to use the separate module.


docker-compose.yml
```diff
@@ -112,6 +112,34 @@ services:
...
+      - GH_TOKEN=${GH_TOKEN:-''}
+      - NPM_TOKEN=${NPM_TOKEN:-''}
+    volumes:
+      - ./scripts/wait-for-it.sh:/wait-for-it.sh
+    links:
+      - mongo
+      - rabbit
+      - bitcoind-1
+      - bitcoind-2
+    command: /wait-for-it.sh -h rabbit -p 5672 -t 30 -- /wait-for-it.sh -h mongo -p 27017 -- npm run start:blockchain-writer
```
lautarodragan: @krobi64 @WesleyCharlesBlake if we're overriding the command here, is the new dockerfile needed at all?

src/Helpers/moduleStartStopCatch.ts
```diff
@@ -0,0 +1,7 @@
+import * as Pino from 'pino'
+
+export const shutdownModule = (server: any) =>
```
lautarodragan: We can avoid this `any` if we declare an interface for `start()` and `stop()` and have the `interface BlockchainWriter` extend from it.

Maybe we could call it [Lifecycle](https://docs.spring.io/spring/docs/3.2.x/javadoc-api/org/springframework/context/Lifecycle.html) or StartStoppable?

```ts
interface Lifecycle {
  readonly start: () => Promise<void>
  readonly stop: () => Promise<void>
}
```

lautarodragan:
This is looking great!

docker-compose.yml
```diff
@@ -112,6 +112,34 @@ services:
       - ipfs
     command: /wait-for-it.sh -h rabbit -p 5672 -t 30 -- /wait-for-it.sh -h mongo -p 27017 -- npm start
 
+  poet-node_blockchain-writer:
```
lautarodragan: Maybe use dashes everywhere for consistency? `poet-node-blockchain-writer`

package.json
```diff
@@ -7,6 +7,7 @@
   },
   "scripts": {
     "start": "node dist/babel/src",
+    "start:blockchain-writer": "node dist/babel/src/BlockchainWriter/index.json",
```
lautarodragan:
```suggestion
    "start:blockchain-writer": "node dist/babel/src/BlockchainWriter/index.js",
```

lautarodragan:
- Tested manually and it's working :rocket: 
- Current test suite runs fine according to Travis :rocket: :rocket:

src/BlockchainWriter/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      batchWriterCreateNextBatchSuccess: configuration.exchangeBatchWriterCreateNextBatchSuccess,
+      poetAnchorDownloaded: configuration.exchangePoetAnchorDownloaded,
+      claimsDownloaded: configuration.exchangeClaimsDownloaded,
+      purgeStaleTransactions: configuration.exchangePurgeStaleTransactions,
+    },
+  })
+}
+
+blockchainWriter().start()
+  .then(shutdownModule)
```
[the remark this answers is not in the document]
lautarodragan: Good catch.

src/Helpers/moduleStartStopCatch.ts
kennylavender: Or it could just take a stop callback instead of the entire object, I can't think of a reason why this function would need to know about the entire object.

Also maybe rename this function, shutdownModule sounds like your commanding it to shut down, but it's really an event handler so maybe something like `onShutdownEvent` or `handleShutdownEvent`. There may be a better name those are just a couple off the top of my head.

docker-compose.yml
WesleyCharlesBlake: Nope. We can override the commands in the deployment manifests. The same approach I used for the regtest miner. So we're good here.

krobi64: So I don't need a separate Dockerfile? Nice.

WesleyCharlesBlake: Correct. Provided I have the commands I can override the Docker file entry point

docker-compose.yml
krobi64: That's fine, I was thinking that the 'lodash' was a way to separate the higher order name from the individual module.

src/Helpers/moduleStartStopCatch.ts
krobi64: I'll play with it. I had an interface defined, but it was complaining at the time. I took it out, but I think it was an unrelated issue, and I didn't get back to it.

PR #937, opened 2019-02-20
patch(Docker): minimal Docker image

WesleyCharlesBlake:

### Release

## Description of Changes
Provides a minimal Docker image based on node-alpine linux.
Adds a posix shell compatible wait-for-it script for better portability with shells that are not bash


Dockerfile
```diff
@@ -1,25 +1,21 @@
...
-                       gcc-5 \
-                       g++-5 \
-                       && rm -rf /var/lib/apt/lists/* \
-                       && rm /etc/apt/sources.list.d/unstable.list
+COPY package*.json /usr/src/app/
+COPY . /usr/src/app/
 
-RUN rm /usr/bin/g++ && ln -s /usr/bin/g++-5 /usr/bin/g++
+RUN apk add --no-cache --virtual .gyp python make git g++ libtool autoconf automake rsync \
+    && npm install
```
lautarodragan: Curious about this:
- Why running `npm install` like this rather than in a separate line?
- Why no longer `npm ci`?

WesleyCharlesBlake: running the installs in one directive reduces the digest layers, making it easier to make a smaller final Docker image. Since the Docker image is a mutli stage build, it makes it a bit simpler. I can move the npm install to its on `RUN` directive and see the outcomes.

`npm ci` is not available in the node:alpine images. I did try get it working but ran into some issues. Let me spend some more time on it this morning

lautarodragan: Got it! I'm fine with running installs in a single directive.

Would prefer `npm ci` over `npm i` though. I think `npm i` may yield different results for the same package.json and package-lock.json, so we may wind up with very slight (but possible annoying or even problematic) differences. But as long as it's working, I think we're fine merging it as it is and figuring it out in a separate PR.

I'm curious though, why isn't it available in alpine images? It comes with `npm`. It is using an old version of npm?

WesleyCharlesBlake: yeah I was stumped as well... oh wait, I think I may have done something wrong previously, npm ci is working now ! :) one second... will push the updates now

